PR #1, opened 2019-01-17
Implement java if statements

krinart:
Example code, verified in JVM runtime:
```java
public class Model {

    public static double score(double[] input) {
        double var0;
        double var1;
        if ((1) == (1)) {
            var1 = 1;
        } else {
            var1 = 2;
        }

        if ((var1 + 2) >= (1 / 2)) {
            var0 = 1;
        } else {
            var0 = 2;
        }

        return var0;
    }

}
```


izeigerman:
Looks awesome 👍

PR #2, opened 2019-01-17
Introduce TransparentExpression which can be omitted during interpretation

krinart:


m2cgen/interpreters/base.py
```diff
@@ -8,20 +8,19 @@ def interpret(self, expr):
...
         handler_name = self._handler_name(type(expr))
         if hasattr(self, handler_name):
             return getattr(self, handler_name)
 
-        for expr_tpe in fallback_tpes:
-            if isinstance(expr, expr_tpe):
-                handler_name = self._handler_name(expr_tpe)
-                if hasattr(self, handler_name):
-                    return getattr(self, handler_name)
-
```
krinart: This code was doing nothing - it's exactly the same as previous block of code.

izeigerman: Noooooooo, don't remove this :) This is a base class and a convenience operation. Considering that we may potentially support third-party interpreters, it's up to a developer to a developer how exactly implement it (whether to handle expressions individually or group them). What do you think? Am I over engineering this?

m2cgen/ast.py
```diff
@@ -107,3 +107,10 @@ def __init__(self, test, body, orelse):
     def __str__(self):
         args = ",".join([str(self.test), str(self.body), str(self.orelse)])
         return "IfExpr(" + args + ")"
+
+
+# Transparent Expressions
+
+class TransparentExpr(Expr):
```
izeigerman: Not sure about the name. It doesn't really tell how to use it, but only suggests how it might be interpreted. Perhaps `LogicalBlockExpr`? Or `LogicalUnitExpr`?  Or even `DiscreteExpr`/`DiscreteUnitExpr`?

krinart:
Deprecated.

PR #5, opened 2019-01-18
Total refactoring: introduce exporters and code_generators

krinart:
1. Move assemblers/interpreters to the new m2cgen.ast module
2. Split interpreter into interpreter and code_generator
3. Introduce exporters

Since now interpreter resides under m2cgen.ast, it is only responsible to interpret ast expression, nothing else. All decorations, such as packages/class/method definitions are done by exporter.

Both exporters and interpreters share SAME instance of code_generator which is actually responsible to properly populate result code.

It brings better separation of responsibilities (exporter vs interpreter, interpreter vs code_generator), allows reusing code_generators by both interpreters and exporters. Will allow to extend exporters with extra logic such as validation.


m2cgen/code_generators/base.py
```diff
@@ -0,0 +1,31 @@
+class BaseCodeGenerator:
+
+    code = None
```
izeigerman: Why it's not instance specific?

PR #7, opened 2019-01-19
Introduce idea of Grammar

krinart:
Experimental PR to play with idea of Grammars.

Overall idea of Grammars seems pretty good as now adding new languages is even more simple and clear. All we need to do is to define basic building blocks of code:

```python
    class JavaGammar(BaseGrammar):
        num_value = "${value}"
        comp_expression = "(${left}) ${op} (${right})"
        bin_num_expression = "(${left}) ${op} (${right})"
        var_declaration = "${var_type} ${var_name};"
        return_statement = "return ${value};"
        array_index_access = "${array_name}[${index}]"
        if_statement = """
    if (${if_def}) {
        ${body_def}
    } else {
        ${else_body}
    }
    """
```

Yet the problem is that now we have following code in BaseCodeGenerator which simply proxies calls to the grammar:

```python
    def comp_expression(self, left, right, op):
        return self.grammar.comp_expression(left=left, right=right, op=op)

    def bin_num_expression(self, left, right, op):
        return self.grammar.bin_num_expression(left=left, right=right, op=op)

    def num_value(self, value):
        return self.grammar.num_value(value=value)

    def array_index_access(self, array_name, index):
        return self.grammar.array_index_access(
            array_name=array_name, index=index)
```


krinart:
Closed on favor of https://github.com/BayesWitnesses/m2cgen/pull/10

PR #14, opened 2019-01-22
Minor fixes for code_generator

krinart:


m2cgen/ast/interpreters/code_generator.py
```diff
@@ -50,13 +48,11 @@ def decrease_indent(self):
         assert self._current_indent >= 0, (
             "Invalid indentation: {}".format(self._current_indent))
 
-    def add_code_line(self, line, newline=True):
-        indent = "".join([" "] * self._current_indent)
-        self.code += indent + line
-        if newline:
-            self.code += "\n"
+    def add_code_line(self, line):
+        indent = "" * self._current_indent
```
izeigerman: I believe this update breaks the indentation logic.

krinart: Good catch! Fixed.

PR #16, opened 2019-01-22
Remove one layer of modules

izeigerman:


m2cgen/interpreters/__init__.py
```diff
@@ -0,0 +1,7 @@
+from .java.interpreter import JavaInterpreter
+from .python.interpreter import PythonInterpreter
+
+__all__ = [
+    JavaInterpreter,
+    PythonInterpreter
+]
```
krinart: Looks like the only place where we could use this code is in exporters.py - and we don't. So I don't think we need this code.

izeigerman: Don't we use it in "exporters.py"?

krinart: My bad!

PR #20, opened 2019-01-23
Add tests for LinearRegressionAssembler

krinart:


tests/utils.py
```diff
@@ -0,0 +1,26 @@
...
+        assert isinstance(left, type(right)), (
+            "Expected instance of {}, received {}".format(
+                type(right), type(left)))
+
+        # Only compare attributes which don't start with __
+        attrs_to_compare = filter(
+            lambda attr_name: not attr_name.startswith('__'), dir(left))
+
+        for attr_name in attrs_to_compare:
+            assert cmp_exprs(
```
izeigerman: Do we need to compare attributes that are not expressions? Like `value` in `NumVal`? Or `index` in `FeatureRef`?

izeigerman: Or `op` in `BinNumOp`

krinart: Why not? Why should we treat them as arbitrary values not worth testing? Those values are part of the contract which assemblers should strictly obey since invalid values mean invalid produced ast.

izeigerman: But don't we have to ensure that the position of coefficients (in case of regression) corresponds to positions of relevant feature values?

PR #23, opened 2019-01-24
Add python e2e tests

krinart:


tests/e2e/executors/java.py
```diff
@@ -19,6 +19,7 @@ def predict(self, X):
         files_to_compile = []
         for model_name, code in self.exporter.export():
             file_name = os.path.join(dirpath, "{}.java".format(model_name))
+            print(file_name)
```
izeigerman: Do we need this print?

krinart: Removed

PR #29, opened 2019-01-24
Improve e2e tests

krinart:
1. By default run executor for all X_test even if some fail and print failed
2. Add --xx flag to fail fast


krinart:
Probably over-engineering.

PR #33, opened 2019-01-28
Some improvements to make code after introduction of for classification unit

krinart:


m2cgen/interpreters/interpreter.py
```diff
@@ -15,26 +15,24 @@ def interpret(self, expr):
     # Default method implementations
 
     def interpret_if_expr(self, expr, if_var_name=None,
-                          is_multi_output=False, **kwargs):
+                          _is_vector_output=False, **kwargs):
```
izeigerman: Considering that we can retrieve `is_vector_output` value from any expression, do we still need this argument?

m2cgen/assemblers/linear.py
```diff
@@ -10,19 +9,17 @@ def assemble(self):
...
-            if coef.shape[0] == 1:
-                return _linear_to_ast(coef[0], intercept[0])
-            else:
-                exprs = []
-                for idx in range(coef.shape[0]):
-                    exprs.append(ast.SubroutineExpr(
-                        _linear_to_ast(coef[idx], intercept[idx])))
-                return ast.ArrayExpr(exprs)
+        coef = utils.to_2d_array(self.model.coef_)
+        intercept = utils.to_2d_array(self.model.intercept_)
```
izeigerman: Why do we need to transform this to 2d array? We always expect either scalar or a 1d array here.

m2cgen/ast.py
```diff
@@ -49,10 +49,10 @@ def __str__(self):
 
 
 class ArrayExpr(NumExpr):
```
izeigerman: Do we need to rename this to `VectorExpr`?

m2cgen/interpreters/java/interpreter.py
```diff
@@ -28,58 +30,52 @@ def interpret(self, expr):
...
+        _is_vector_output = subroutine.expr._is_vector_output
+
         self._cg = self._create_code_generator()
 
         with self._cg.method_definition(
                 name=subroutine.name,
-                args=[("double[]", self._feature_array_name)],
-                return_type=return_type):
+                args=[
+                    (True, self._feature_array_name)],
```
izeigerman: I'm not sure I like the way `args` parameter looks now. Before it represented type -> value. I'm not sure it generalizes well. Are we ever going to generate methods with arguments other than `double` and `double[]`?

m2cgen/interpreters/java/code_generator.py
```diff
@@ -31,13 +33,16 @@ def class_definition(self, model_name):
         self.add_block_termination()
 
     @contextlib.contextmanager
-    def method_definition(self, name, args, return_type, modifier="public"):
+    def method_definition(self, name, args, return_vector, modifier="public"):
```
izeigerman: `return_vector`? I'm not sure I understand what does that mean. We might return non-vector value as well.

izeigerman: Ah, it's boolean. Let's rename it to something that would tell so.

m2cgen/interpreters/code_generator.py
```diff
@@ -119,3 +130,6 @@ class CLikeCodeGenerator(BaseCodeGenerator):
     tpl_else_statement = CodeTemplate("} else {")
     tpl_block_termination = CodeTemplate("}")
     tpl_var_assignment = CodeTemplate("${var_name} = ${value};")
+
+    scalar_variable_type = "double"
```
izeigerman: `variable` is a bad name, since we use this for method return type as well. How about `output`?

m2cgen/interpreters/java/code_generator.py
```diff
@@ -15,7 +15,9 @@ def add_class_def(self, class_name, modifier="public"):
 
     def add_method_def(self, name, args, return_type, modifier="public"):
         method_def = modifier + " static " + return_type + " " + name + "("
-        method_def += ",".join([t + " " + n for t, n in args])
+        method_def += ",".join([
```
izeigerman: It's a bit confusing that we still need to figure out the var type for arguments in this method, but `return_type` already contains a type itself.

m2cgen/ast.py
krinart: Done

m2cgen/interpreters/code_generator.py
krinart: Done

m2cgen/interpreters/java/code_generator.py
krinart: Done

m2cgen/assemblers/linear.py
krinart: Fixed

PR #36, opened 2019-01-29
RandomForestClassifier for python

krinart:


m2cgen/ast.py
```diff
@@ -48,18 +48,53 @@ def __str__(self):
...
 
 
-class VectorExpr(NumExpr):
+# Vector Expressions.
+
+class VectorExpr(Expr):
     is_vector_output = True
 
+
+class VectorVal(VectorExpr):
```
izeigerman: Can we call it `VectorValExpr` to maintain consistency with other AST members.

m2cgen/interpreters/interpreter.py
```diff
@@ -5,6 +5,9 @@
 
 class BaseInterpreter:
 
+    with_vectors = False
+    with_linear_algebra = False
```
izeigerman: Aren't these 2 attributes interchangeable? I can't find a use case when we may have one without another.

m2cgen/interpreters/python/interpreter.py
```diff
@@ -17,6 +19,23 @@ def interpret(self, expr):
             last_result = self._do_interpret(expr)
             self._cg.add_return_statement(last_result)
 
+        if self.with_linear_algebra or self.with_vectors:
+            self._cg.prepend_code_line("import numpy as np")
```
izeigerman: Code leak into interpreter? I thought we should delegate all code related logic to Code Generator instead.

m2cgen/interpreters/interpreter.py
krinart: Done

m2cgen/ast.py
krinart: Ack

izeigerman:
🎉 🥇 👍

PR #40, opened 2019-01-30
Implement regression for C

krinart:


m2cgen/interpreters/c/code_generator.py
```diff
@@ -0,0 +1,26 @@
...
+
+class CCodeGenerator(CLikeCodeGenerator):
+
+    def __init__(self, *args, **kwargs):
+        super(CCodeGenerator, self).__init__(*args, **kwargs)
+
+    def add_function_def(self, name, args, is_vector_output):
+        return_type = self._get_var_type(is_vector_output)
+
+        method_def = return_type + " " + name + "("
```
izeigerman: So is this function? Or a method? 😆

m2cgen/interpreters/c/code_generator.py
```diff
@@ -0,0 +1,26 @@
...
+        method_def += ",".join([
+            self._get_var_type(False) + " " + n + "[]"
```
izeigerman: This looks like a workaround. Instead I suggest to just override `vector_output_type` with `double *`. Actually we should remove `vector_output_type` from `CLikeCodeGenerator` and make specific ones for Java and C.

izeigerman:
Fantastic 👍 🥇

PR #44, opened 2019-01-31
Implement classification for C

krinart:


m2cgen/interpreters/c/code_generator.py
```diff
@@ -4,10 +4,20 @@
 from m2cgen.interpreters.code_generator import CodeTemplate as CT
 
 
+ARRAY_ASSIGNMENT_FUNC = """
+void assignArray(double *output, double input[], int size) {
```
izeigerman: Snake case is a lot more common in C. Let's make it `assign_array`.

krinart: Done

m2cgen/assemblers/__init__.py
```diff
@@ -7,3 +7,24 @@
...
+    "LinearRegression": LinearModelAssembler,
+    "LogisticRegression": LinearModelAssembler,
+    "DecisionTreeRegressor": TreeModelAssembler,
+    "DecisionTreeClassifier": TreeModelAssembler,
+    "RandomForestRegressor": RandomForestModelAssembler,
+    "RandomForestClassifier": RandomForestModelAssembler,
+}
+
+
+def get_assembler_cls(model):
```
izeigerman: Why it was moved here?

m2cgen/interpreters/code_generator.py
```diff
@@ -16,6 +16,8 @@ def __call__(self, *args, **kwargs):
 
 class BaseCodeGenerator:
 
+    initial_code = ""
```
izeigerman: I don't think this code deserves to be in a BaseGenerator, considering that it's way to language-specific + we do initial generation in the top level `interpret` method of the corresponding interpreter. Also I'm not sure I understand why `assign_array` needs to be "initial" since we need to add it only in case of classification.

m2cgen/interpreters/c/code_generator.py
```diff
@@ -45,6 +52,21 @@ def add_var_declaration(self, expr):
         self.add_code_line(tpl(var_name=var_name, size=size))
         return var_name
 
+    def add_var_assignment(self, var_name, value, expr):
```
izeigerman: The method signature has become obscure. It's no longer clear what arguments are for. We have `var_name` and `value`, so what is `expr` and what's its role in variable assignment? It's just a bit misleading.

tests/e2e/executors/Executor.java
```diff
@@ -25,9 +25,7 @@ public static void main(String[] args) {
                 String out = "";
                 for (int i = 0, l = arr.length; i < l; i++) {
                     out += arr[i];
-                    if (i < l - 1) {
```
izeigerman: Why this has been removed? Now it'll output the unnecessary trailing whitespace.

m2cgen/interpreters/code_generator.py
krinart: Agreed, fixed.

m2cgen/interpreters/c/code_generator.py
krinart: Done :) thanks!

PR #45, opened 2019-01-31
Implement simple CLI; add Python 3.7; Improve setup.py; Split travis build into unit/e2e runs

krinart:


m2cgen/cli.py
```diff
@@ -0,0 +1,50 @@
...
+    "java": m2cgen.export_to_java,
+}
+
+
+parser = argparse.ArgumentParser(
+    description="Generate code in native language for provided model")
+parser.add_argument("infile", type=argparse.FileType("rb"), nargs="?",
+                    default=sys.stdin.buffer,
+                    help="File with pickle representation of the model")
+parser.add_argument("--language", dest="language", type=str,
```
izeigerman: Looks great, but I'd say that the `package_name` is pretty important parameter in case of Java. Indentation can be added latter.

setup.py
```diff
@@ -1,21 +1,38 @@
...
+    keywords="sklearn ml code-generation",
+    python_requires="!=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*",
+    install_requires=[
+        "numpy",
+        "scipy",
+        "scikit-learn",
+    ],
+    entry_points={
+        'console_scripts': [
+            'm2c = m2cgen.cli:main',
```
izeigerman: let's use the full project name for the CLI (like `python -m m2cgen`)

m2cgen/cli.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def generate_code(args):
+    with args.infile as f:
+        model = pickle.load(f)
+
+    exporter, supported_args = LANGUAGE_TO_EXPORTER[args.language]
+
+    kwargs = {}
+    for arg_name in supported_args:
```
izeigerman: Shall we check that the language was specified and invoke `parser.print_usage()` if not?

m2cgen/cli.py
```diff
@@ -0,0 +1,74 @@
...
+    help="Target language",
+    required=True)
+parser.add_argument(
+    "--class_name", "-cn", dest="class_name", type=str,
+    help="Name of the generated class (if supported by target language)")
+parser.add_argument(
+    "--package_name", "-pn", dest="package_name", type=str,
+    help="Package name for the generated code "
+         "(if supported by target language)")
+parser.add_argument(
```
izeigerman: `--help` is missing.

m2cgen/cli.py
izeigerman: Shall we handle invalid language?

setup.py
izeigerman: Single quotes?

izeigerman: Btw, does `python -m m2cgen` work for you?

PR #57, opened 2019-02-04
Limit the depth of the binary numeric expression in Python

izeigerman:


m2cgen/interpreters/interpreter.py
```diff
@@ -1,18 +1,67 @@
 import re
+import sys
 
 from m2cgen import ast
 
 
-class BaseInterpreter:
+class BaseAstInterpreter:
```
izeigerman: I really don't like `Ast` prefix here. It's kind of obvious what's being interpreted. Same true for `AstToCode...` why not just `CodeGenInterpreter`? Or even just `ToCodeInterpreter`?

tests/interpreters/test_python.py
```diff
@@ -169,3 +169,69 @@ def  score(input):
...
+def test_depth_threshold_with_bin_expr():
+    expr = ast.NumVal(1)
+    for i in range(4):
+        expr = ast.BinNumExpr(ast.NumVal(1), expr, ast.BinNumOpType.ADD)
+
+    interpreter = CustomPythonInterpreter()
+
+    expected_code = """
+def  score(input):
+    var0 = (1) + ((1) + (1))
```
izeigerman: Can we add a test case where `BinExpr` are diluted with some other exprs?
Like `BinNumExpr` -> `BinNumExpr` -> `SubroutineExpr`-> `BinNumExpr` -> `IfExpr` -> `BinNumExpr`. Something along those lines.

krinart:
Closed in favor of #60

PR #62, opened 2019-02-07
XGBoost support

izeigerman:


m2cgen/interpreters/interpreter.py
```diff
@@ -119,3 +100,33 @@ def interpret_vector_val(self, expr, **kwargs):
...
+            if expr not in self._cached_exprs:
+                result = handler(expr, **kwargs)
+                var_name = self._cg.add_var_declaration(expr.output_size)
+                self._cg.add_var_assignment(var_name, result, expr.output_size)
+                self._cached_exprs[expr] = var_name
+                return var_name
+            else:
+                return self._cached_exprs[expr]
+        else:
+            return handler(expr, **kwargs)
```
krinart: I think this can be simplified a little:

```python
if not expr.to_cache:
    return handler(...)

if expr in self._cached_exprs:
    return self._cached_exprs[expr]

...
```

m2cgen/assemblers/utils.py
```diff
@@ -57,3 +57,26 @@ def to_2d_array(var):
...
+
+
+def sigmoid_expr(expr):
+    neg_expr = ast.BinNumExpr(ast.NumVal(0), expr, ast.BinNumOpType.SUB)
+    exp_expr = ast.ExpExpr(neg_expr)
+    return ast.BinNumExpr(
+        ast.NumVal(1),
+        ast.BinNumExpr(ast.NumVal(1), exp_expr, ast.BinNumOpType.ADD),
+        ast.BinNumOpType.DIV,
+        to_reuse=True)
```
krinart: I think `to_reuse` should be function argument

m2cgen/assemblers/utils.py
```diff
@@ -57,3 +57,26 @@ def to_2d_array(var):
...
+
+
+def softmax_exprs(exprs):
+    exp_exprs = [ast.ExpExpr(e, to_reuse=True) for e in exprs]
+    exp_sum_expr = ast.BinNumExpr(
+        exp_exprs[0],
+        apply_op_to_expressions(ast.BinNumOpType.ADD, *exp_exprs[1:]),
+        ast.BinNumOpType.ADD,
+        to_reuse=True)
```
krinart: It looks like you can replace all of this with something like:
```python
exp_sum_expr = apply_op_to_expressions(ast.BinNumOpType.ADD, exp_exprs, to_reuse=True)
```

m2cgen/assemblers/utils.py
```diff
@@ -57,3 +57,26 @@ def to_2d_array(var):
...
+    exp_expr = ast.ExpExpr(neg_expr)
+    return ast.BinNumExpr(
+        ast.NumVal(1),
+        ast.BinNumExpr(ast.NumVal(1), exp_expr, ast.BinNumOpType.ADD),
+        ast.BinNumOpType.DIV,
+        to_reuse=True)
+
+
+def softmax_exprs(exprs):
+    exp_exprs = [ast.ExpExpr(e, to_reuse=True) for e in exprs]
```
krinart: I'm not sure I understand why we need `to_reuse` for these expressions as they seem to only used once.

m2cgen/assemblers/xgboost.py
```diff
@@ -0,0 +1,106 @@
...
+
+class XGBoostModelAssembler(ModelAssembler):
+
+    def __init__(self, model):
+        super().__init__(model)
+        self._base_score = self.model.base_score
+        self._n_estimators = self.model.n_estimators
+
+        feature_names = enumerate(self.model.get_booster().feature_names)
+        self._feature_name_to_idx = {name: idx for idx, name in feature_names}
```
krinart: Nit: Usually `enumerate` is used in the loop, but I will leave it it's up to you.

m2cgen/assemblers/xgboost.py
```diff
@@ -0,0 +1,106 @@
...
+
+        use_lt_comp = tree["missing"] == tree["no"]
+        if use_lt_comp:
+            comp_op = ast.CompOpType.LT
+            true_child_id = tree["yes"]
+            false_child_id = tree["no"]
+        else:
+            comp_op = ast.CompOpType.GTE
+            true_child_id = tree["no"]
+            false_child_id = tree["yes"]
```
krinart: I have a gut feeling that we can get rid of this conditional block altogether, as we just switch operation and true/false blocks. Something like:
```python
comp_op = tree["missing"] == tree["no"] and ast.CompOpType.LT else ast.CompOpType.GTE
true_child_id = tree["yes"]
false_child_id = tree["no"]
```

m2cgen/assemblers/xgboost.py
```diff
@@ -0,0 +1,106 @@
...
+
+        return ast.IfExpr(ast.CompExpr(feature_ref, threshold, comp_op),
+                          self._assemble_child_tree(tree, true_child_id),
+                          self._assemble_child_tree(tree, false_child_id))
+
+    def _assemble_child_tree(self, tree, child_id):
+        for child in tree["children"]:
+            if child["nodeid"] == child_id:
+                return self._assemble_tree(child)
+        assert False, "Unexpected child ID {}".format(child_id)
```
krinart: nit: Maybe raising `ValueError` would be more appropriate? `assert False` just gives me a strong feeling of forgotten debugging code :)

m2cgen/assemblers/xgboost.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def _split_trees_by_classes(trees, n_classes):
+    # Splits are computed based on a comment
+    # https://github.com/dmlc/xgboost/issues/1746#issuecomment-267400592.
+    trees_by_classes = [[] for _ in range(n_classes)]
+    for i in range(len(trees)):
+        class_idx = i % n_classes
+        trees_by_classes[class_idx].append(trees[i])
```
krinart:
```python
trees_by_classes = defaultdict(list)
for i, tree in enumerate(trees):
    trees_by_classes[i].append(tree)
```

m2cgen/ast.py
```diff
@@ -3,6 +3,7 @@
 
 class Expr:
     output_size = 1
+    to_reuse = False
```
krinart: Maybe one line comment what `to_reuse` means?

m2cgen/interpreters/interpreter.py
```diff
@@ -71,7 +49,10 @@ class ToCodeInterpreter(BaseToCodeInterpreter):
     about AST.
     """
 
-    with_vectors = False
+    def __init__(self, cg, feature_array_name="input"):
+        super().__init__(cg, feature_array_name=feature_array_name)
+        self.with_vectors = False
+        self._reused_exprs = {}
```
krinart: nit: Maybe call it `cached_expr_results` or something? `_reused_exprs` gives me a feeling of list or set with somehow actually reused expressions.

m2cgen/assemblers/xgboost.py
izeigerman: Well, we need assert here because we just verify our assumption. This assert should never (never!) fire. ValueError on the other hand assumes that we do some sort of validation and throw an error which can be handled. There is nothing we can do if this condition is False though.

krinart: Ok, makes sense.

m2cgen/assemblers/utils.py
izeigerman: Nope. The used twice: once to calculate some and the second time - for normalization (`exp(val[i]) / sum`

m2cgen/assemblers/utils.py
izeigerman: Not really. I want to `reuse` only last expression and not all of them. For example in expression:
```
(x1 +(x2 + (x3)))
```
I want only to cache the whole thing, and not each subexpression.

m2cgen/assemblers/xgboost.py
izeigerman: Not sure I understand. Isn't this a loop?

m2cgen/assemblers/xgboost.py
izeigerman: Have you noticed that in one case true == "yes" and in other case true is "no"?

krinart: Exactly.
First case gives us:
```python
if (<):
  yes
else:
  no
```

Second case gives us:
```python
if (>):
  no
else:
  yes
```

We can have:
```python
if (<op>):
  yes
else:
  no
```

With `op` depending on the context.

Unless the actual order of `yes/no` matters, of course.

m2cgen/assemblers/xgboost.py
krinart:
```python
feature_names = self.model.get_booster().feature_names
self._feature_name_to_idx = {
    name: idx for idx, name in enumerate(feature_names)}
```

Otherwise you need to remeber that `feature_names` is already result of `enumerate`.

m2cgen/assemblers/utils.py
krinart: Right. If you apply `to_reuse` only to the result (and not recursively to all operations) - you will get exactly this.

m2cgen/assemblers/xgboost.py
```diff
@@ -0,0 +1,106 @@
...
+        model_dump = self.model.get_booster().get_dump(dump_format="json")
+        trees = [json.loads(d) for d in model_dump]
+
+        if self._is_classification:
+            if self._output_size == 1:
+                return self._assemble_bin_class_output(trees)
+            else:
+                return self._assemble_multi_class_output(trees)
+        else:
+            return self._assemble_single_output(trees, self._base_score)
```
krinart: For some reason I am very sensitive to unnecessary nested blocks.

Maybe again make them flat:

```python
if not self._is_classification:
    return self._assemble_single_output(trees, self._base_score)

if self._output_size == 1:
    return self._assemble_bin_class_output(trees)

return self._assemble_multi_class_output(trees)

```

izeigerman: Well, I have an exact opposite sensitivity 😆 Not a big fun of premature return statement which are not last statements in the flow. This leads to messy return points in the code which are hard to keep track of. Seems like writing code in expressions only had irreversible impact on me.

m2cgen/assemblers/utils.py
izeigerman: Ah, I see. Good point.

m2cgen/assemblers/xgboost.py
izeigerman: I'm not sure how this
```
true_child_id = tree["yes"]
false_child_id = tree["no"]
```
can possible give me this:
```
if (>):
  no
else:
  yes
```

izeigerman: Besides I'm not even sure how to read this statement:
```
comp_op = tree["missing"] == tree["no"] and ast.CompOpType.LT or ast.CompOpType.GTE
```
What does `and` even mean here? Neither `ast.CompOpType.LT` nor `ast.CompOpType.GTE` are boolean expressions.

m2cgen/assemblers/xgboost.py
izeigerman: On the other hand branching gives you expression-like code with clearly defined logic flow.

m2cgen/interpreters/interpreter.py
izeigerman: Fixed

m2cgen/ast.py
izeigerman: Fixed

m2cgen/assemblers/xgboost.py
izeigerman: Fixed

m2cgen/assemblers/utils.py
izeigerman: Fixed

m2cgen/assemblers/xgboost.py
```diff
@@ -0,0 +1,108 @@
...
+    def _assemble_tree(self, tree):
+        if "leaf" in tree:
+            return ast.NumVal(tree["leaf"])
+
+        threshold = ast.NumVal(tree["split_condition"])
+        feature_idx = self._feature_name_to_idx[tree["split"]]
+        feature_ref = ast.FeatureRef(feature_idx)
+
+        use_lt_comp = tree["missing"] == tree["no"]
+        if use_lt_comp:
```
krinart: Maybe leave a comment why we are doing this? As just looking at this code it's not clear why bother with 2 versions when they seem to be the same.

m2cgen/interpreters/interpreter.py
```diff
@@ -9,12 +9,13 @@ class BaseInterpreter:
...
     """
+    def __init__(self):
+        self._cached_expr_results = {}
 
     def interpret(self, expr):
+        self._reset_reused_expr_cache()
         return self._do_interpret(expr)
 
-    # Private methods implementing Visitor pattern
-
```
krinart: Was this comment intentionally deleted? I think it adds some clarity about what exactly we are doing here.

m2cgen/interpreters/mixins.py
```diff
@@ -121,6 +121,7 @@ def process_subroutine_queue(self, top_code_generator):
         self._subroutine_idx = 0
 
         while len(self.subroutine_expr_queue):
+            self._reset_reused_expr_cache()
```
krinart: Nice! 👍

tests/e2e/test_e2e.py
```diff
@@ -62,8 +65,13 @@ def classification_binary(model):
...
     # previous list.
     [
+        # XGBoost
+        regression(xgboost.XGBRegressor(**XGBOOST_PARAMS)),
+        classification(xgboost.XGBClassifier(**XGBOOST_PARAMS)),
+        classification_binary(xgboost.XGBClassifier(**XGBOOST_PARAMS)),
+
         # SVM
-        # regression(svm.LinearSVR(random_state=RANDOM_SEED)),
+        regression(svm.LinearSVR(random_state=RANDOM_SEED)),
```
krinart: Oops.. thanks you!

m2cgen/interpreters/interpreter.py
izeigerman: nope, accident

m2cgen/assemblers/xgboost.py
izeigerman: Fixed

m2cgen/assemblers/xgboost.py
```diff
@@ -0,0 +1,111 @@
...
+from m2cgen.assemblers import utils
+from m2cgen.assemblers.base import ModelAssembler
+
+
+class XGBoostModelAssembler(ModelAssembler):
+
+    def __init__(self, model):
+        super().__init__(model)
+        self._base_score = self.model.base_score
+        self._n_estimators = self.model.n_estimators
```
krinart: I think this one is not used.

izeigerman: It's not, thanks!

izeigerman: Fixed

PR #65, opened 2019-02-11
LightGBM support

krinart:


README.md
```diff
@@ -51,8 +51,8 @@ pip install m2cgen
       </tr>
       <tr>
         <th>Boosting</th>
-        <td>XGBClassifier(gbtree/dart booster only)</td>
-        <td>XGBRegressor(gbtree/dart booster only)</td>
+        <td>XGBClassifier(gbtree/dart booster only), LGBMClassifier</td>
```
izeigerman: Please mention in parenthesis that we support gbdt/dart boosters only for LGBM as well.

m2cgen/assemblers/boosting.py
```diff
@@ -5,36 +5,40 @@
 from m2cgen.assemblers.base import ModelAssembler
 
 
-class XGBoostModelAssembler(ModelAssembler):
+class BaseBoostAssembler(ModelAssembler):
```
izeigerman: `BaseBoostingAssembler`?

m2cgen/assemblers/boosting.py
```diff
@@ -5,36 +5,40 @@
...
+        super().__init__(model)
+        self.trees = trees
+        self._base_score = base_score
 
         self._output_size = 1
         self._is_classification = False
+
         model_class_name = type(model).__name__
-        if model_class_name == "XGBClassifier":
+        if model_class_name == self.classifier_name:
```
izeigerman: We already have constructor arguments, why not pass the `classifier_name` there?

m2cgen/assemblers/boosting.py
```diff
@@ -5,36 +5,40 @@
...
-                return self._assemble_bin_class_output(trees)
+                return self._assemble_bin_class_output(self.trees)
             else:
-                return self._assemble_multi_class_output(trees)
+                return self._assemble_multi_class_output(self.trees)
         else:
-            return self._assemble_single_output(trees, self._base_score)
+            return self._assemble_single_output(self.trees, self._base_score)
+
+    def _assemble_single_output(self, trees, base_score=0):
```
izeigerman: We already have a `trees` attribute in `self` so it looks quite confusing to pass it in arguments as well. Can we rename the `self` one to `all_trees` or something?

m2cgen/assemblers/boosting.py
```diff
@@ -60,13 +67,24 @@ def _assemble_bin_class_output(self, trees):
...
         ])
 
-    def _assemble_single_output(self, trees, base_score):
-        trees_ast = [self._assemble_tree(t) for t in trees]
-        result_ast = utils.apply_op_to_expressions(
-            ast.BinNumOpType.ADD,
-            ast.NumVal(base_score),
-            *trees_ast)
-        return ast.SubroutineExpr(result_ast)
+    def _assemble_tree(self, node):
```
izeigerman: So do we assemble tree? Or node? :) Let's rename the argument.

m2cgen/ast.py
```diff
@@ -141,6 +141,17 @@ class CompOpType(Enum):
     EQ = '=='
     NOT_EQ = '!='
 
+    @staticmethod
+    def from_str_op(op):
+        return {
+            ">": CompOpType.GT,
```
izeigerman: I think this can be slightly improved:
```
class CompOpType(Enum):
    GT = '>'
    GTE = '>='
    LT = '<'
    LTE = '<='
    EQ = '=='
    NOT_EQ = '!='

    @staticmethod
    def from_str_op(op):
        return COMP_OP_TYPE_MAPPING[op]


COMP_OP_TYPE_MAPPING = {e.value: e.name for e in CompOpType}
```

m2cgen/assemblers/boosting.py
krinart: Done

README.md
krinart: Done

m2cgen/assemblers/boosting.py
krinart: Done

m2cgen/assemblers/boosting.py
krinart: Done

m2cgen/ast.py
krinart: Thanks!

m2cgen/assemblers/boosting.py
krinart: The thing is that this parameter is not instance-specific, but rather class-specific. So I don't think it should be a constructor parameter.

m2cgen/assemblers/boosting.py
```diff
@@ -100,6 +119,33 @@ def _assemble_child_tree(self, tree, child_id):
...
+    classifier_name = "LGBMClassifier"
+
+    def __init__(self, model):
+        model_dump = model.booster_.dump_model()
+        trees = model_dump["tree_info"]
+
+        super().__init__(model, trees)
+
+    def _assemble_tree(self, tree):
+        if "tree_structure" in tree:
```
izeigerman: Apparently each `tree` object should container either `tree_structure` or `leaf_value` otherwise something went very wrong. Can we express this with the `assert`?

krinart: Done, thanks!

README.md
izeigerman: `gbtree` is called `gbdt` in lightGBM :)

PR #70, opened 2019-03-05
fix typo

Alisa-lisa:
The usage example is for python

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ pip install m2cgen
 
 ## Usage
 
-Here's a simple example of how a trained linear model can be represented in Java code:
+Here's a simple example of how a trained linear model can be represented in Python code:
 ```python
 from sklearn.datasets import load_boston
 from sklearn import linear_model


izeigerman:
Hi @Alisa-lisa ! Thank you so much for submitting this PR! The idea of that example was to demonstrate how using `m2cgen` API once can generate **Java** code from the trained linear regression model. This is why it says `Java` there. Perhaps this whole `Usage` section can be rewritten somehow to make it clearer and avoid this kind of confusion?

Alisa-lisa:
Ah, the positioning was a bit misleading for me+)
Maybe breaking one-line phrase into 2 would be more understandable for noobs like me+)
like:
Here's a simple example of how a trained linear in model in one language can be represented in a different language.
Model training in Python:
{code}
Generated Java code:
{code}

izeigerman:
@Alisa-lisa I really like your proposal! What if we specify that model training happens in Python specifically, since this the only way of model training `m2cgen` currently supports. Like:

Here's a simple example of how a linear model trained in Python environment can be represented in Java code:
{code}
Generated Java code:
{code}

Alisa-lisa:
You mean: 
1. Show training in python
2. Equal code in Java
3. Generated code from python to Java?

Sounds perfect!

izeigerman:
Hey @Alisa-lisa are you planning to finish this update or should I close this PR?

izeigerman:
Closing this PR due to lack of activity.

PR #78, opened 2019-03-19
Feat: Add Dockerfile, Makefile

matbur:
I had a problem with installing `lightgbm` on mac OS, so I think this PR will help anyone to run this app on all operating systems.


izeigerman:
@matbur thank you so much for your contribution! This update should indeed make development process much easier. I have several questions though:
1. What is the motivation behind using `pipenv` inside of Docker? Isn't it a bit overkill? Basically it's a virtual environment inside of a virtual environment. Can we just use a combination of `setup.py` and `requirements-test.txt` instead? My concern is that with `pipenv` we introduce yet another place where the list of dependencies must be maintained which makes the whole process a bit cumbersome and complex.
2. I saw you added JDK to your image 👍 In our E2E tests we also rely on `gcc` to test generated C code. What versions of `gcc` and `libc` are used in that image?

matbur:
Thanks @izeigerman,
1. I used `pipenv` because I don't want to force anyone to use `docker` and I prefer using `pipenv` to run tests on local machine. If you want, I can use simple `requirements-test.txt`.
2. Versions are:
```
$ docker run --rm python:3.7 /lib/x86_64-linux-gnu/libc.so.6
```
```
GNU C Library (Debian GLIBC 2.24-11+deb9u4) stable release version 2.24, by Roland McGrath et al.
Compiled by GNU CC version 6.3.0 20170516.
```

krinart:
:+1: for not using `pipenv` as it bring unnecessary complexity.

izeigerman:
@matbur let's not introduce `pipenv` for now. Perhaps going forward we can consider using it instead of our current `requirements-test.txt` setup. 
Meanwhile if I'm not mistaken we can do something like this in the Dockerfile:
```
COPY requirements-test.txt ./

RUN pip install -r ./requirements-test.txt

CMD python setup.py develop && pytest -v -x --fast
```
@krinart @matbur does the snippet above look correct?

Additionally I suggest to consider using JDK from the default package manager instead of relying on direct external links without any mirrors:
```
RUN apt-get update && apt-get install -y openjdk-8-jdk
ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
```
Once updated I think this PR is good to go.

PR #82, opened 2019-03-27
Add ntree-limit fix for XGBoostModelAssembler

arshamg:
Currently the XGBoostModelAssembler ignores ntree_limit in the XGBoost model. This differs with the sklearn implementation of XGBoost models. This fix respects the ntree_limit (if exists) when building the transpiled model.

### DIFF
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -84,7 +84,7 @@ class XGBoostModelAssembler(BaseBoostingAssembler):
 
         model_dump = model.get_booster().get_dump(dump_format="json")
 
-        # Respect sklearn ntree_limit
+        # Respect XGBoost ntree_limit
         ntree_limit = getattr(model, "best_ntree_limit", 0)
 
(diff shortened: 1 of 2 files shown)


m2cgen/assemblers/boosting.py
```diff
@@ -83,6 +83,13 @@ def __init__(self, model):
         }
 
         model_dump = model.get_booster().get_dump(dump_format="json")
+
+        # Respect sklearn ntree_limit
```
izeigerman: I think it's not part of `sklearn` but rather `XGBoost`, right?

izeigerman:
@arshamg thank you so much for contributing this fix! This is indeed a quite significant overlook on our side. Like @krinart pointed out can you please create a simple unit test that covers this scenario.

arshamg:
Okay great - I will do!

arshamg:
Added a rudimentary test - not sure if this is the best way.

krinart:
@arshamg Thanks a lot!

PR #83, opened 2019-04-02
Fix application of best_ntree_limit to the entire list of estimators.  Instead the limit is applied to per-class estimators split

izeigerman:


m2cgen/assemblers/boosting.py
```diff
@@ -34,6 +36,10 @@ def assemble(self):
                 self.all_trees, self._base_score)
 
     def _assemble_single_output(self, trees, base_score=0):
+        if self._tree_limit is not None:
+            assert self._tree_limit > 0, "Unexpected tree limit"
```
krinart: I think `__init__` method would be better place for this assert:
- No need to perform it every time this function runs
- It would raise an error right when we pass it, not when we use it later

izeigerman: Good point 👍

izeigerman: Fixed

PR #86, opened 2019-04-22
 Initial implementation of the SVM model assembler

izeigerman:
Please note that the multi-class support as well as generated examples will be added in a subsequent PR.

### DIFF
--- a/m2cgen/assemblers/svm.py
+++ b/m2cgen/assemblers/svm.py
@@ -1,0 +1,97 @@
+from m2cgen import ast
+from m2cgen.assemblers import utils
+from m2cgen.assemblers.base import ModelAssembler
+
+
+class SVMModelAssembler(ModelAssembler):
+
+    def __init__(self, model):
+        super().__init__(model)
+
+        supported_kernels = {
+            "rbf": self._rbf_kernel,
+            "sigmoid": self._sigmoid_kernel,
+            "poly": self._poly_kernel,
+            "linear": self._linear_kernel
+        }
+        kernel_type = model.kernel
+        if kernel_type not in supported_kernels:
+            raise ValueError("Unsupported kernel type {}".format(kernel_type))
+        self._kernel_fun = supported_kernels[kernel_type]
+
+        n_features = len(model.support_vectors_[0])
+
+        self._gamma = model.gamma
+        if self._gamma == "auto" or self._gamma == "auto_deprecated":
+            self._gamma = 1.0 / n_features
+
+        self._output_size = 1
+        if type(model).__name__ == "SVC":
+            n_classes = len(model.n_support_)
+            if n_classes > 2:
+                self._output_size = n_classes
+
+    def assemble(self):
+        return self._assemble_single_output()
+
+    def _assemble_single_output(self):
+        support_vectors = self.model.support_vectors_
+        coef = self.model.dual_coef_[0]
+        intercept = self.model.intercept_[0]
+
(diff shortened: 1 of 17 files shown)


m2cgen/assemblers/svm.py
```diff
@@ -0,0 +1,97 @@
...
+
+        self._gamma = model.gamma
+        if self._gamma == "auto" or self._gamma == "auto_deprecated":
+            self._gamma = 1.0 / n_features
+
+        self._output_size = 1
+        if type(model).__name__ == "SVC":
+            n_classes = len(model.n_support_)
+            if n_classes > 2:
+                self._output_size = n_classes
```
izeigerman: This will be used for multi-class in the follow-up PR.

m2cgen/assemblers/svm.py
```diff
@@ -0,0 +1,97 @@
...
+        negative_gamma = utils.sub(ast.NumVal(0), ast.NumVal(self._gamma))
+        kernel = None
+        for j in range(len(support_vector)):
+            sub_expr = utils.sub(ast.NumVal(support_vector[j]),
+                                 ast.FeatureRef(j))
+            pow_expr = ast.PowExpr(sub_expr, ast.NumVal(2))
+            if kernel:
+                kernel = utils.add(pow_expr, kernel)
+            else:
+                kernel = pow_expr
```
krinart: It's up to you, but I think using `utils.apply_op_to_expressions` (see my comment to `_linear_kernel` function) makes code simpler.

m2cgen/assemblers/svm.py
```diff
@@ -0,0 +1,97 @@
...
+    def _linear_kernel(self, support_vector):
+        kernel = None
+        for j in range(len(support_vector)):
+            mul_expr = utils.mul(ast.NumVal(support_vector[j]),
+                                 ast.FeatureRef(j))
+            if kernel:
+                kernel = utils.add(mul_expr, kernel)
+            else:
+                kernel = mul_expr
+        return kernel
```
krinart:
```python
return utils.apply_op_to_expressions(
    utils.mul(ast.NumVal(support_el), ast.FeatureRef(i)
    for i, support_el in enumerate(support_vector), 
    ast.BinNumOpType.ADD)
```
Does it make sense?

m2cgen/interpreters/interpreter.py
```diff
@@ -89,9 +89,14 @@ class ToCodeInterpreter(BaseToCodeInterpreter):
     about AST.
     """
 
+    exponent_function_name = NotImplemented
+    power_function_name = NotImplemented
+    tanh_function_name = NotImplemented
```
krinart: `NotImplemented` is an alias to the `None`, and while in the code it looks good, in runtime we would get a native code which attempts to call function called `None`.

The best thing I can think of is just adding an assert to verify that the corresponding function is not empty.

izeigerman: Will do. I just followed the same logic as in https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/interpreters/code_generator.py#L26-L34

izeigerman: Fixed

m2cgen/assemblers/svm.py
izeigerman: Fixed

m2cgen/assemblers/svm.py
izeigerman: Fixed

PR #93, opened 2019-07-17
Fixed missing feature names for XGBoost

akhvorov:
Now it's not fail if XGBoost model were saved by save without feature names

### DIFF
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -84,9 +84,11 @@ class XGBoostModelAssembler(BaseBoostingAssembler):
 
     def __init__(self, model):
         feature_names = model.get_booster().feature_names
-        self._feature_name_to_idx = {
-            name: idx for idx, name in enumerate(feature_names)
-        }
+        self._feature_name_to_idx = {}
+        if feature_names is not None:
+            self._feature_name_to_idx = {
+                name: idx for idx, name in enumerate(feature_names)
+            }
 
         model_dump = model.get_booster().get_dump(dump_format="json")
         trees = [json.loads(d) for d in model_dump]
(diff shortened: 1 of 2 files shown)


izeigerman:
Hey @akhvorov, thank you so much for your PR! The fix is really great 👍 May I please ask you to add a unit test to make sure that the issue is gone? Otherwise I can add one myself in the follow-up PR.

m2cgen/assemblers/boosting.py
```diff
@@ -84,9 +84,11 @@ class XGBoostModelAssembler(BaseBoostingAssembler):
 
     def __init__(self, model):
         feature_names = model.get_booster().feature_names
```
krinart: Thanks for this fix!
Maybe it would be easier to make
```python
feature_names = model.get_booster().feature_names or []
```
Or maybe even better
```python
name: idx for idx, name in enumerate(feature_names or [])
```
?

tests/assemblers/test_xgboost.py
```diff
@@ -228,3 +228,47 @@ def test_multi_class_best_ntree_limit():
...
+def test_regression_saved_without_feature_names():
+    import os
+
+    base_score = 0.6
+    estimator = xgboost.XGBRegressor(n_estimators=2, random_state=1,
+                                     max_depth=1, base_score=base_score)
+    utils.train_model_regression(estimator)
+
+    filename = "tmp.file"
+    if os.path.exists(filename):
```
izeigerman: Thanks for adding the test! 
May I please ask you to use `tmp_dir` utility function to avoid polluting user's workspace. Example can be found here:
https://github.com/BayesWitnesses/m2cgen/blob/00abbff1e978536a642355932a1a9c121feea8ca/tests/e2e/executors/base.py#L12
Note that this context manager will clean up the temporary directory automatically so you don't have to do it manually here. 
I'd also suggest to use UUID for a filename in place of hardcoded `tmp.file` to avoid potential conflicts, since this particular test should not be affected by artifacts which can potentially be produced by other tests.

Great job otherwise 👍

krinart:
Thanks for the contribution 👍

akhvorov:
Thank you for advices and help!

PR #97, opened 2019-09-10
Adding Javascript interpreter to support scoring in BigQuery.

bcampbell-prosper:


bcampbell-prosper:
I'm hoping to get the Javascript interpreter we're using to do scoring in BigQuery merged into the main repo. Let me know what you guys think or if there are any changes you'd like me to make/tests I missed, etc...

Sorry about the plethora of commits, I'm pretty new to Python development and didn't have flake8 or pytest set up in my environment.

I love this package BTW!

izeigerman:
Hey @bcampbell-prosper! Your contribution is highly appreciated - it would be awesome to have Javascript support in **m2cgen**!  Please note that it'd take me a few days to complete the review. Fantastic job 👍

m2cgen/interpreters/javascript/utils.js
```diff
@@ -0,0 +1,15 @@
...
+    return input;
+}
+
+function dict_to_value_array(input) {
+    var output = [];
+    for (const elem in input) {
+        output.push(input[elem]);
+    }
+    return output;
+}
```
krinart: What is the purpose of these functions? They do not seem to be used by `JavascriptCodeGenerator`.

izeigerman: I second this. What's the idea behind the `with_util_functions` parameter?

m2cgen/interpreters/javascript/interpreter.py
```diff
@@ -0,0 +1,61 @@
...
+from m2cgen.interpreters import mixins
+from m2cgen.interpreters import utils
+from m2cgen.interpreters.interpreter import ToCodeInterpreter
+from m2cgen.interpreters.javascript.code_generator \
+    import JavascriptCodeGenerator
+
+
+class JavascriptInterpreter(ToCodeInterpreter,
+                            mixins.LinearAlgebraMixin,
+                            mixins.SubroutinesAsFunctionsMixin):
```
izeigerman: Are you sure that JS code will benefit from splitting the generated code into subroutines? The initial purpose of splitting code into smaller functions was to workaround limitations related to the size of an individual function/method. Eg. JVM has strict constrains on how large the individual method can be - this is why we use subroutines for Java code. 

If there are no similar constrains in JS and if having smaller functions doesn't improve the runtime, I'd suggest to remove subroutine support to make the implementation simpler.

m2cgen/interpreters/javascript/utils.js
bcampbell-prosper: The util functions are mostly used within BigQuery.  I was on the fence about adding them in as they aren't used by M2CGEN itself but they make passing inputs in much easier.  BigQuery has a TO_JSON_STRING() function that takes a row as input making this very easy to pass JSON to a JavaScript function.  Once it's in JavaScript we always need to convert `null` to `NaN` and if we choose to use the default array input we need to convert the dict to an array.  We can do the conversion ahead of time but it's much easier with the util functions. Also, the way we're using the JavaScript models is first converting the code then string search/replacing `input[0]` to `input.featureA`.

bcampbell-prosper: If you'd like me to exclude these and the corresponding parameter let me know.

m2cgen/interpreters/javascript/interpreter.py
bcampbell-prosper: I'm glad to remove the subroutine support.  I honestly wasn't sure what `SubroutinesAsFunctions` was doing and just copied the Java implementation to start.

I'll add the e2e tests, and remove the `SubroutinesAsFunctionsMixin`.  Let me know what you want done with the util functions.

m2cgen/interpreters/javascript/utils.js
izeigerman: @bcampbell-prosper I'd say these functions are quite specific to the BigQuery use case and are outside of the `m2cgen`'s domain. This is why I suggest to remove them. 

I believe it shouldn't be too hard to include them in your own build process right after the generated code is produced. You can, for instance, append those functions to the generated code string.

bcampbell-prosper: Makes sense.  I'll take them out.

bcampbell-prosper:
Awesome! Glad to contribute.

Issue #99, opened 2019-09-27
Support for LightGBM Booster


chris-smith-zocdoc:
We're training our LightGBM model outside of python (spark) so we need to load it from a model file before passing it to m2c. I don't believe LightGBM can load directly into `LGBMRegressor` though, it must be loaded into lgb.Booster. 

It would be nice if m2cgen supported lgb.Booster 

Example
```py
import lightgbm as lgb
import m2cgen as m2c

model = lgb.Booster(model_file='model.txt')

# this fails
# m2c.export_to_java(model)

# This works but is awkward 
from lightgbm.sklearn import LGBMRegressor
r = LGBMRegressor()
r._Booster = model

code = m2c.export_to_java(r)
```

izeigerman:
Hey @chris-smith-zocdoc, thanks for reporting this! I think the direct `Booster` support is worth adding to `m2cgen`. As part of this effort I'd also suggest to add a direct `Booster` instance support for `XGBoost` models as well. 

Btw, PR is very welcome if you're up to it :)

chris-smith-zocdoc:
I can give it a shot, can you point me to the appropriate files that would need changed?

izeigerman:
Thanks, @chris-smith-zocdoc! You can begin with the following lines:
https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/assemblers/boosting.py#L139 - for LightGBM
https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/assemblers/boosting.py#L86 - for XGBoost.

This is where we're accessing the underlying `Booster` instances from `scikit-learn` compatible wrappers. I believe we can try and check what's being passed to us - a wrapper or a `Booster` instance, and if it's a wrapper - retrieve the underlying `Booster` instance from it.

Jie-Yuan:
classifier don't work

alexeymaksakov-tomtom:
Also for LGBMRegressor I observe issues woth operations type:
```
model_name = data_path + "LightGBM_model1.txt"
model = lgb.Booster(model_file=model_name)

from lightgbm.sklearn import LGBMRegressor
r = LGBMRegressor()
r._Booster = model

code = m2c.export_to_java(r)```

results in
```  File "/home/maksakov/workspace/search-scoring-optimization/venv/lib/python3.8/site-packages/m2cgen/assemblers/boosting.py", line 123, in <listcomp>
    return [self._assemble_tree(t) for t in trees]
[...]

PR #104, opened 2019-10-14
Codegen for LightGBM Booster

chris-smith-zocdoc:
Allows generating code from `lightgbm.Booster` Fixes https://github.com/BayesWitnesses/m2cgen/issues/99

I tried doing the same for XGBoost but wasn't able to find the right apis to read the necessary attributes (objective function, num classes). Maybe someone with more XGBoost familiarity could take this on.

### DIFF
--- a/m2cgen/assemblers/__init__.py
+++ b/m2cgen/assemblers/__init__.py
@@ -16,6 +16,7 @@ SUPPORTED_MODELS = {
     # LightGBM
     "LGBMRegressor": LightGBMModelAssembler,
     "LGBMClassifier": LightGBMModelAssembler,
+    "Booster": LightGBMModelAssembler,
 
     # XGBoost
     "XGBClassifier": XGBoostModelAssembler,
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -1,5 +1,6 @@
 import json
 import numpy as np
+import lightgbm
 from m2cgen import ast
 from m2cgen.assemblers import utils
 from m2cgen.assemblers.base import ModelAssembler
@@ -9,22 +10,16 @@ class BaseBoostingAssembler(ModelAssembler):
 
     classifier_name = None
 
-    def __init__(self, model, trees, base_score=0, tree_limit=None):
+    def __init__(self, model, trees, base_score=0, tree_limit=None, output_size=1, is_classification=False):
         super().__init__(model)
         self.all_trees = trees
         self._base_score = base_score
 
-        self._output_size = 1
-        self._is_classification = False
+        self._output_size = output_size
+        self._is_classification = is_classification
 
         assert tree_limit is None or tree_limit > 0, "Unexpected tree limit"
         self._tree_limit = tree_limit
-
-        model_class_name = type(model).__name__
-        if model_class_name == self.classifier_name:
-            self._is_classification = True
-            if model.n_classes_ > 2:
-                self._output_size = model.n_classes_
 
     def assemble(self):
         if self._is_classification:
@@ -95,8 +90,18 @@ class XGBoostModelAssembler(BaseBoostingAssembler):
         # assembling (if applicable).
         best_ntree_limit = getattr(model, "best_ntree_limit", None)
 
+        model_class_name = type(model).__name__
+        is_classification = False
+        output_size = 1
+
+        if model_class_name == self.classifier_name:
+            is_classification = True
+            if model.n_classes_ > 2:
+                output_size = model.n_classes_
+
         super().__init__(model, trees, base_score=model.base_score,
-                         tree_limit=best_ntree_limit)
+                         tree_limit=best_ntree_limit, output_size=output_size, 
+                         is_classification=is_classification)
 
     def _assemble_tree(self, tree):
         if "leaf" in tree:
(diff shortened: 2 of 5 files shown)


tests/assemblers/test_lightgbm.py
```diff
@@ -110,3 +111,73 @@ def test_regression():
...
+def test_regression_from_booster():
+    estimator = lightgbm.LGBMRegressor(n_estimators=2, random_state=1, max_depth=1)
+    utils.train_model_regression(estimator)
+
+    with utils.tmp_dir() as tmp_dirpath:
+        filename = os.path.join(tmp_dirpath, "model.txt")
+        estimator.booster_.save_model(filename)
+        booster = lightgbm.Booster(model_file=filename)
+
+    assembler = assemblers.LightGBMModelAssembler(booster)
```
chris-smith-zocdoc: These are the same tests as above except we save the booster to a file, then use the assembler on the booster instead of the SKLearn Model

m2cgen/assemblers/boosting.py
```diff
@@ -134,12 +139,30 @@ def _assemble_child_tree(self, tree, child_id):
...
+            if 'objective' not in model_dump:
+                raise Exception("objective function not specified in model_dump. try upgrading lightgbm >= 2.3.0")
+
+            objective = model_dump['objective']
+            is_classification = 'multiclass' in objective or 'binary' in objective
+            
+            if output_size > 2:
+                output_size = output_size
+
+        super().__init__(model, trees, output_size=output_size, is_classification=is_classification)
```
krinart: I wonder how this code is different from the following one:
```python
    if isinstance(model, lightgbm.Booster):
        model_dump = model.dump_model()
    else:
        model_dump = model.booster_.dump_model()
    trees = [m["tree_structure"] for m in model_dump["tree_info"]]
    super().__init__(model, trees)
```
Given that the base constructor remains intact?

tests/assemblers/test_lightgbm.py
```diff
@@ -110,3 +111,73 @@ def test_regression():
...
+
+
+def test_regression_from_booster():
+    estimator = lightgbm.LGBMRegressor(n_estimators=2, random_state=1, max_depth=1)
+    utils.train_model_regression(estimator)
+
+    with utils.tmp_dir() as tmp_dirpath:
+        filename = os.path.join(tmp_dirpath, "model.txt")
+        estimator.booster_.save_model(filename)
+        booster = lightgbm.Booster(model_file=filename)
```
krinart: nit: there's a `model_str` argument which potentially allows not to use a temporary file (in conjunction with `Booster.dump_model` method). I might be wrong however as I didn't verify it myself.

m2cgen/assemblers/__init__.py
```diff
@@ -16,6 +16,7 @@
     # LightGBM
     "LGBMRegressor": LightGBMModelAssembler,
     "LGBMClassifier": LightGBMModelAssembler,
+    "Booster": LightGBMModelAssembler,
```
izeigerman: I'm not sure this will work out nicely since `XGBoost` use the same class name for its Booster.

m2cgen/assemblers/boosting.py
izeigerman: I believe this is not enough since `type(model).__name__` is no longer a reliable source of truth about whether this is a classifier or regressor.

m2cgen/assemblers/boosting.py
```diff
@@ -9,23 +10,17 @@ class BaseBoostingAssembler(ModelAssembler):
 
     classifier_name = None
```
izeigerman: Do we still need this? Can we remove it? For `LightGBMModelAssembler` and `XGBoostModelAssembler` classes as well.

m2cgen/assemblers/boosting.py
```diff
@@ -134,12 +139,30 @@ def _assemble_child_tree(self, tree, child_id):
 class LightGBMModelAssembler(BaseBoostingAssembler):
 
     classifier_name = "LGBMClassifier"
-
+
```
izeigerman: [Cosmetic] Unnecessary indentation here.

tests/assemblers/test_lightgbm.py
izeigerman: Can we somehow extract expected ASTs into separate functions/variables to avoid extensive code duplication?

m2cgen/assemblers/boosting.py
krinart: Ahhh... got it. So now `model` variable which pass to base constructor sometimes will be of type `LGBMRegressor`/`LGBMClassifier` and sometimes of type `Booster`. This makes me think something is fundamentally wrong here. 

Although one possible workaround could be start treating `trees` itself as our model. Any thoughts?

m2cgen/assemblers/__init__.py
krinart: This is another point which makes me question whether we really need this change. 

@chris-smith-zocdoc just to double check, all these changes are required to simply avoid 2 following lines of code:
```
estimator = lgb.LGBMClassifier()
estimator._Booster = booster
```

Is it correct?

m2cgen/assemblers/boosting.py
chris-smith-zocdoc: I don't think `model` ever ends up being used, I'd prefer to remove it entirely. 

If we pass trees around I'm not sure how much reuse we'd get as LightGBM and XGBoost have different properties

m2cgen/assemblers/__init__.py
chris-smith-zocdoc: There are some cases where more properties need to be set 

If `base_score` is used, you'll also need to set `estimator.base_score`

For LGBMClassifier set `estimator.n_classes_`


Maybe it makes sense to have a function to convert Booster -> LGBMRegressor/LGBMClassifier instead?

krinart:
Hey @chris-smith-zocdoc 

First of all thanks a lot for your effort - it is truly amazing! 

However this PR contains certain changes which I and @izeigerman have some concerns about. 

Specifically:

1. Using `Booster` class name as a marker for deciding which assembler to use. As @izeigerman pointed out, `xgboost` has [the same name](https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L1048) for booster class, so this mechanism is not going to work smoothly for both `lightgbm` and `xgboost` boosters.

2. Passing different types as a `model` to `BaseBoostingAssembler` class in different cases: sometimes estimator and sometimes booster. Imagine if it was a statically typed language such as Java - we couldn't do such trick. Just because we can do it in Python doesn't mean we should.

I think having a converter function from booster to estimator is the only viable option we have right now. However in order to solve the problem of boosters having the same class name, the client should use the appropriate function, something like this:

```python
import m2cgen as m2c

booster = ...
m2c.export_to_java(m2c.xgbooster_to_estimator(booster))
```

I'm open to other ideas solving those 2 issues.

Regards,
Victor

chris-smith-zocdoc:
At this moment we can't support XGBoost Booster afaict because it doesn't expose enough information to recreate the sklearn estimator.

I can add `m2c.lightgbm_booster_to_estimator(booster)` if you want it explicitly named or just `m2c.booster_to_estimator(booster)`. For the second option I can inspect `__module__` in addition to `__name__` verify only lgm booster is passed in for now.

image

chris-smith-zocdoc:
I closed this as I'm no longer working on it, some else is free to take this over if they wish

Issue #103, opened 2019-10-01
Large LightGBM causes javac error "Code too Large"


chris-smith-zocdoc:
When generating code for a large number of trees, the generated code exceeds the 64KB limit in java. 

From [Stackoverflow](https://stackoverflow.com/questions/2407912/code-too-large-compilation-error-in-java)
> A single method in a Java class may be at most 64KB of bytecode.

One solution is to add subfunctions https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/assemblers/boosting.py#L43-L48 instead of having the body of every tree inside `subroutine0`. The amount of code that will fit inside each function is dependent on its depth + width so we might require some heuristic or tunable parameter. In my case, I ended up with 10 trees per subfunction

I'm not sure if there are similar limits in other languages

izeigerman:
Hey @chris-smith-zocdoc! 

This is rather weird because we're aware of this Java limitations and this is why we came up with subroutines in the first place. I remember testing this implementation with as many as 500 - 1000 estimators with XGBoost, LightGBM and Random Forest without any problem. 

Can you please provide some steps to reproduce this issue locally? Ideally using some public dataset or the one available in the scikit-learn package?

chris-smith-zocdoc:
I think the issue is that the trees are not in separate subroutines 
![image](https://user-images.githubusercontent.com/2537809/65980281-d68d3c00-e444-11e9-8aef-428183604270.png)

To reproduce
```python
import lightgbm as lgb
import m2cgen as m2c
import numpy as np


N = 10000
np.random.seed(seed=7)
data = np.random.random(size=(N, 200))
target = np.random.random(size=(N, 1))


estimator = lgb.LGBMRegressor(n_estimators=100, random_state=1, max_depth=64, num_leaves=100)

estimator.fit(data, target)

res = m2c.export_to_java(estimator)


with open('Model.java', 'w') as f:
    f.write(res)
```

Then

```
javac Model.java
```

izeigerman:
Ah, I see. So the individual tree is pretty large. Ok, we may want to consider to wrap individual estimators in their own subroutines based on some threshold values for `max_depth` and `num_leaves`.

MilanCugur:
Hi, the same situation happens with large (sklearns) decision trees. 

Do you have a solution for that (is it planned if it doesn't exist?)

izeigerman:
[...]

PR #105, opened 2019-10-14
Limit the number of leaves in each subroutine for gradient boosted trees

chris-smith-zocdoc:
Fixes https://github.com/BayesWitnesses/m2cgen/issues/103


tests/utils.py
```diff
@@ -120,8 +151,13 @@ def verify_python_model_is_expected(model_code, input, expected_output):
 
 
 def predict_from_commandline(exec_args):
-    result = subprocess.Popen(exec_args, stdout=subprocess.PIPE)
-    items = result.stdout.read().decode("utf-8").strip().split(" ")
+    result = subprocess.Popen(exec_args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    stdout, stderr = result.communicate()
+    if result.returncode is not 0:
```
chris-smith-zocdoc: This now handles failure and log the error if the exit code is non-zero

tests/utils.py
```diff
@@ -74,9 +74,40 @@ def train_model_classification_binary(estimator, test_fraction=0.1):
     return _train_model(estimator, datasets.load_breast_cancer(),
                         test_fraction)
 
+def train_model_regression_random_data(estimator, test_fraction=0.01):
```
chris-smith-zocdoc: added a couple method to make testing with synthetic data easier

tests/e2e/test_e2e_large_trees.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+# Absolute tolerance. Used in np.isclose to compare 2 values.
+# We compare 6 decimal digits.
+ATOL = 1.e-6
+
+RANDOM_SEED = 1234
+XGBOOST_PARAMS = dict(base_score=0.6, n_estimators=100, max_depth=12,
+                      random_state=RANDOM_SEED)
+LIGHT_GBM_PARAMS = dict(n_estimators=100, num_leaves=100, max_depth=64,
```
chris-smith-zocdoc: This is copy of the `test_e2e` file except we focus on training large trees

m2cgen/assemblers/boosting.py
```diff
@@ -19,6 +20,7 @@ def __init__(self, model, trees, base_score=0, tree_limit=None):
 
         assert tree_limit is None or tree_limit > 0, "Unexpected tree limit"
         self._tree_limit = tree_limit
+        self._total_num_leaves = total_num_leaves
```
krinart: Will this value properly represent number fo leaves in case when we use `_assemble_multi_class_output` method? I have a feeling that we need to calculate number of leaves for every tree we pass into `_assemble_single_output` separately rather than having it as a class-level attribute.

tests/e2e/test_e2e_large_trees.py
```diff
@@ -0,0 +1,107 @@
...
+        CLASSIFICATION,
+    )
+
+
+def classification_binary(model):
+    return (
+        model,
+        utils.train_model_classification_binary_random_data,
+        CLASSIFICATION,
+    )
```
krinart: Maybe move it to `e2e/__init__.py` module as this is common for all e2e tests?

tests/e2e/test_e2e_large_trees.py
```diff
@@ -0,0 +1,107 @@
...
+@utils.cartesian_e2e_params(
+    # These are the languages which support all models specified in the
+    # next list.
+    [
+        (executors.PythonExecutor, PYTHON),
+        (executors.JavaExecutor, JAVA),
+        (executors.CExecutor, C),
+        (executors.GoExecutor, GO),
+        (executors.JavascriptExecutor, JAVASCRIPT),
+    ],
```
krinart: I think this also can be factored out into `e2e/__init__.py` as a `ALL_EXECUTORS` or something like that.

tests/e2e/test_e2e_large_trees.py
```diff
@@ -0,0 +1,107 @@
...
+
+    idxs_to_test = [0] if is_fast else range(len(X_test))
+
+    with executor.prepare_then_cleanup():
+        for idx in idxs_to_test:
+            y_pred_executed = executor.predict(X_test[idx])
+            print("expected={}, actual={}".format(y_pred_true[idx],
+                                                  y_pred_executed))
+            res = np.isclose(y_pred_true[idx], y_pred_executed, atol=ATOL)
+            assert res if isinstance(res, bool) else res.all()
```
krinart: This seems to be a complete duplication and potentially can be abstracted into `execute_single_e2e_test`.

m2cgen/assemblers/boosting.py
```diff
@@ -4,12 +4,13 @@
 from m2cgen.assemblers import utils
 from m2cgen.assemblers.base import ModelAssembler
 
+MAX_LEAVES_PER_SUBROUTINE = 2000
```
krinart: Would be great if you could add some considerations about this value. Maybe what are the values for estimators/depth you tested it against.

tests/e2e/test_e2e_large_trees.py
chris-smith-zocdoc: I can combine this entire file with `test_e2e.py`

m2cgen/assemblers/boosting.py
chris-smith-zocdoc: Will fix

m2cgen/assemblers/boosting.py
chris-smith-zocdoc: I've tested against a bunch of parameters and haven't found a satisfying answer here, the amount of code generated seems to depend not only on the number of leaves but also the number of estimators. 

Searching of the parameter space, here are the breaking points for the given parameters:

```
| total_num_leaves | n_estimators | max_leaves |
|------------------|--------------|------------|
| 6448             | 10           | 3300       |
| 11125            | 20           | 2800       |
| 14344            | 30           | 2400       |
| 10986            | 50           | 2900       |
| 16650            | 100          | 2200       |
| 16908            | 150          | 1500       |
| 16933            | 175          | 1300       |
| 16958            | 200          | 1200       |
| 17008            | 250          | 900        |
| 17058            | 300          | 800        |
```

izeigerman:
@krinart @chris-smith-zocdoc Considering that having a threshold still doesn't help with some parameter combinations (@chris-smith-zocdoc please correct me if I'm wrong) I wonder whether we should solve this problem on Java interpreter side instead. 

We already addressed a similar issue in Python interpreter by limiting the depth of binary expressions: https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/interpreters/mixins.py#L8-L41. Can we potentially do the same trick for Java and limit the depth of `if` expressions or perhaps even of all expressions?

What do you think?

StrikerRUS:
> Can we potentially do the same trick for Java and limit the depth of if expressions or perhaps even of all expressions?

Sorry for interfering you guys. But I'd like to say that some limits inside a code generator will be very useful for different languages. For example, #109 is stuck due to that we need to limit line length, number of variables and number of continuously following line breakage symbols at the same time for VBA. Also, for some languages a size of code file and number of lines are crucial.

I think it's related.

chris-smith-zocdoc:
> Considering that having a threshold still doesn't help with some parameter combinations

I believe the issue is actually in my current implementation. It doesn't work well if the trees contain drastically different amounts of leaves. Since I'm averaging the leaves per tree it under-estimates how large some trees are, which causes individual subroutines to contain more leaves than the threshold.

A couple alternative implementations:
- very simplistic, cap the number of trees per subroutines to `4`. This number works for very deep trees, but will create extra subroutines for shallow trees

- compute the exact number of leaves per tree before adding it to the current subroutine. This should avoid the averaging issue and isn't very complicated to implement


With both very deep trees and many estimators, we actually run into a new error `Model.java:1: error: too many constants` which is more complicated to solve. I don't have a real model that produces this yet, just my stress test case.


My preference would be to try implementing the "exact leaves per tree" approach before trying the approach you suggested @izeigerman

chris-smith-zocdoc:
@izeigerman I've updated the implementation to count the number of leaves in each tree instead of averaging. This allows it to correctly enforce the limit per subroutine.

I tested a number of different configurations and found the max num leaves to be ~3500. I set it to 3000 as a precaution.

m2cgen/assemblers/boosting.py
```diff
@@ -41,12 +42,47 @@ def _assemble_single_output(self, trees, base_score=0):
...
+        trees_num_leaves = [self._count_leaves(t) for t in trees] 
+
+        if sum(trees_num_leaves) > MAX_LEAVES_PER_SUBROUTINE:
+            to_sum = []
+
+            subroutine_trees = []
+            subroutine_sum_leaves = 0
+            for tree, num_leaves in zip(trees_ast, trees_num_leaves):
+
+                if(len(subroutine_trees) != 0 and subroutine_sum_leaves + num_leaves > MAX_LEAVES_PER_SUBROUTINE):
```
izeigerman: Redundant external parenthesis

m2cgen/assemblers/boosting.py
```diff
@@ -41,12 +42,47 @@ def _assemble_single_output(self, trees, base_score=0):
...
+                    
+                    to_sum.append(ast.SubroutineExpr(partial_result))
+
+                    subroutine_trees = []
+                    subroutine_sum_leaves = 0
+                
+                subroutine_sum_leaves += num_leaves
+                subroutine_trees.append(tree)
+
+            if(len(subroutine_trees) != 0):
```
izeigerman: Ditto: Redundant parenthesis

m2cgen/assemblers/boosting.py
```diff
@@ -130,6 +170,21 @@ def _assemble_child_tree(self, tree, child_id):
...
+    @staticmethod
+    def _count_leaves(tree):
+        queue = [tree]
+        num_leaves = 0
+
+        while queue:
+            tree = queue.pop()
+            if "leaf" in tree:
+                num_leaves += 1
+            elif 'children' in tree:
```
izeigerman: Let's be consistent and use double quotes here

m2cgen/assemblers/boosting.py
```diff
@@ -166,6 +221,20 @@ def _assemble_tree(self, tree):
...
+        queue = [tree]
+        num_leaves = 0
+
+        while queue:
+            tree = queue.pop()
+            if "leaf_value" in tree:
+                num_leaves += 1
+            else:
+                queue.append(tree['left_child'])
+                queue.append(tree['right_child'])
```
izeigerman: Let's be consistent and use double quotes here

izeigerman:
@chris-smith-zocdoc thanks for reiterating on this! I think now it looks a lot better 👍 Let's fix the coverage and style issues and I think we should merge this.

chris-smith-zocdoc:
Thanks @izeigerman , I'm ooo for a week, will fix it up when I get back

izeigerman:
@chris-smith-zocdoc no rush, take your time. Thanks for fixing this in the first place 👍

chris-smith-zocdoc:
I think this is ready for review again

izeigerman:
@chris-smith-zocdoc awesome, thank you! I'll take a look shortly

m2cgen/assemblers/boosting.py
```diff
@@ -4,6 +4,8 @@
 from m2cgen.assemblers import utils
 from m2cgen.assemblers.base import ModelAssembler
 
+MAX_LEAVES_PER_METHOD = 3000
```
izeigerman: I actually liked the `MAX_LEAVES_PER_SUBROUTINE` more, since method is something specific to Java when assemblers should be language independent.

izeigerman:
One last comment and let's merge it.

m2cgen/assemblers/boosting.py
chris-smith-zocdoc: I renamed it because pylint was mad about line lengths (which are super short at 80chars) :(

izeigerman: I'm fine calling it `LEAVES_CUTOFF_THRESHOLD` or ``LEAVES_CUTOFF_LIMIT` if that helps

izeigerman:
@chris-smith-zocdoc I think this fix is quite valuable and I'd love to include it into the next release. Are you planning to finish this PR or should I pick it up from here?

chris-smith-zocdoc:
> @chris-smith-zocdoc I think this fix is quite valuable and I'd love to include it into the next release. Are you planning to finish this PR or should I pick it up from here?

Hey @izeigerman I'm traveling this/next week. I can get to it when I'm back or you can finish it up. You should have permissions to push commits to this branch 
image

izeigerman:
@chris-smith-zocdoc Thank you and please don't worry. There is no rush. I just wanted to check whether this PR has been abandoned or not. Thanks for your reply

izeigerman:
Closing this PR in favor of https://github.com/BayesWitnesses/m2cgen/pull/123

PR #106, opened 2019-10-16
enhance gitignore file

StrikerRUS:
Refer to https://github.com/github/gitignore/blob/master/Python.gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,129 @@
+# macOS
+**/.DS_Store
+
+# IDEs stuff
 .idea
-**/__pycache__
-**/*.pyc
-venv
+.vscode
+
+# Compiled Object files
+*.slo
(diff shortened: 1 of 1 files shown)


izeigerman:
Hi @StrikerRUS! Thanks for your PR. One question though: do we really need all those things in our gitignore? I don't mind I guess but it just seems a bit redundant.

StrikerRUS:
@izeigerman This is standard gitignore file for all Python projects. Moreover, I've removed some things from it, because it's very unlikely they will be here in the future (e.g. `Django`, `Flask`, etc.). As we cannot know in what environment contributors will code, what linters they may use, what way they install packages, we need all this stuff for their convenience.

izeigerman:
Have no objections. Merging

PR #111, opened 2019-10-19
drop numpy dependency from runtime where possible

### LINKED ISSUE #108: drop numpy dependency from Python code for cases without vectors
According to this line, it seems that `numpy` is used as a default math library for runtime even when we do not operate with vectors.
https://github.com/BayesWitnesses/m2cgen/blob/2475f3cddb5b328c8673795ca3cbe4fdc89f6797/m2cgen/interpreters/python/interpreter.py#L30-L31

Let me describe two advantages of dropping `numpy` where it's possible.

The first one is **excess dependence**. Even though `numpy` is a sort of "classic" dependence and there should be no problems with installing it, it requires additional manipulation from a user side. Also, there are some companies with very strict security policies, which prohibit using pip (conda, brew, and other package managers). So, I guess, for them raw Python may be preferable solution in cases where it's possible.

The second one is **speed**. `numpy` is about efficient **vector** math, in other cases it only produces redundant computational cost. Consider the following example. Take [this generated Python code from the repo](https://github.com/BayesWitnesses/m2cgen/blob/master/generated_code_examples/python/classification/svm.py), change return type from `np.array` to simple `list`, replace the following things in script:
- `numpy` -> `math`
- `np.exp` -> `math.exp`
- `np.power` -> `math.pow`

Here what we get after removing `numpy`:
```
import math
def score_raw(input):
    var0 = (0) - (0.25)
    var1 = math.exp((var0) * ((((math.pow((5.4) - (input[0]), 2)) + (math.pow((3.0) - (input[1]), 2))) + (math.pow((4.5) - (input[2]), 2))) + (math.pow((1.5) - (input[3]), 2))))
    var2 = math.exp((var0) * ((((math.pow((6.2) - (input[0]), 2)) + (math.pow((2.2) - (input[1]), 2))) + (math.pow((4.5) - (input[2]), 2))) + (math.pow((1.5) - (input[3]), 2))))
    var3 = math.exp((var0) * ((((math.pow((5.0) - (input[0]), 2)) + (math.pow((2.3) - (input[1]), 2))) + (math.pow((3.3) - (input[2]), 2))) + (math.pow((1.0) - (input[3]), 2))))
    var4 = math.exp((var0) * ((((math.pow((5.9) - (input[0]), 2)) + (math.pow((3.2) - (input[1]), 2))) + (math.pow((4.8) - (input[2]), 2))) + (math.pow((1.8) - (input[3]), 2))))
    var5 = math.exp((var0) * ((((math.pow((5.0) - (input[0]), 2)) + (math.pow((2.0) - (input[1]), 2))) + (math.pow((3.5) - (input[2]), 2))) + (math.pow((1.0) - (input[3]), 2))))
    var6 = math.exp((var0) * ((((math.pow((6.7) - (input[0]), 2)) + (math.pow((3.0) - (input[1]), 2))) + (math.pow((5.0) - (input[2]), 2))) + (math.pow((1.7) - (input[3]), 2))))
    var7 = math.exp((var0) * ((((math.pow((7.0) - (input[0]), 2)) + (math.pow((3.2) - (input[1]), 2))) + (math.pow((4.7) - (input[2]), 2))) + (math.pow((1.4) - (input[3]), 2))))
    var8 = math.exp((var0) * ((((math.pow((4.9) - (input[0]), 2)) + (math.pow((2.4) - (input[1]), 2))) + (math.pow((3.3) - (input[2]), 2))) + (math.pow((1.0) - (input[3]), 2))))
    var9 = math.exp((var0) * ((((math.pow((6.3) - (input[0]), 2)) + (math.pow((2.5) - (input[1]), 2))) + (math.pow((4.9) - (input[2]), 2))) + (math.pow((1.5) - (input[3]), 2))))
    var10 = math.exp((var0) * ((((math.pow((6.0) - (input[0]), 2)) + (math.pow((2.7) - (input[1]), 2))) + (math.pow((5.1) - (input[2]), 2))) + (math.pow((1.6) - (input[3]), 2))))
    var11 = math.exp((var0) * ((((math.pow((5.7) - (input[0]), 2)) + (math.pow((2.6) - (input[1]), 2))) + (math.pow((3.5) - (input[2]), 2))) + (math.pow((1.0) - (input[3]), 2))))
    var12 = math.exp((var0) * ((((math.pow((5.1) - (input[0]), 2)) + (math.pow((3.8) - (input[1]), 2))) + (math.pow((1.9) - (input[2]), 2))) + (math.pow((0.4) - (input[3]), 2))))
    var13 = math.exp((var0) * ((((math.pow((4.4) - (input[0]), 2)) + (math.pow((2.9) - (input[1]), 2))) + (math.pow((1.4) - (input[2]), 2))) + (math.pow((0.2) - (input[3]), 2))))
    var14 = math.exp((var0) * ((((math.pow((5.7) - (input[0]), 2)) + (math.pow((4.4) - (input[1]), 2))) + (math.pow((1.5) - (input[2]), 2))) + (math.pow((0.4) - (input[3]), 2))))
    var15 = math.exp((var0) * ((((math.pow((5.8) - (input[0]), 2)) + (math.pow((4.0) - (input[1]), 2))) + (math.pow((1.2) - (input[2]), 2))) + (math.pow((0.2) - (input[3]), 2))))
    var16 = math.exp((var0) * ((((math.pow((5.1) - (input[0]), 2)) + (math.pow((3.3) - (input[1]), 2))) + (math.pow((1.7) - (input[2]), 2))) + (math.pow((0.5) - (input[3]), 2))))
    var17 = math.exp((var0) * ((((math.pow((5.7) - (input[0]), 2)) + (math.pow((3.8) - (input[1]), 2))) + (math.pow((1.7) - (input[2]), 2))) + (math.pow((0.3) - (input[3]), 2))))
    var18 = math.exp((var0) * ((((math.pow((4.3) - (input[0]), 2)) + (math.pow((3.0) - (input[1]), 2))) + (math.pow((1.1) - (input[2]), 2))) + (math.pow((0.1) - (input[3]), 2))))
    var19 = math.exp((var0) * ((((math.pow((4.5) - (input[0]), 2)) + (math.pow((2.3) - (input[1]), 2))) + (math.pow((1.3) - (input[2]), 2))) + (math.pow((0.3) - (input[3]), 2))))
    var20 = math.exp((var0) * ((((math.pow((6.3) - (input[0]), 2)) + (math.pow((2.7) - (input[1]), 2))) + (math.pow((4.9) - (input[2]), 2))) + (math.pow((1.8) - (input[3]), 2))))
    var21 = math.exp((var0) * ((((math.pow((6.0) - (input[0]), 2)) + (math.pow((3.0) - (input[1]), 2))) + (math.pow((4.8) - (input[2]), 2))) + (math.pow((1.8) - (input[3]), 2))))
    var22 = math.exp((var0) * ((((math.pow((6.3) - (input[0]), 2)) + (math.pow((2.8) - (input[1]), 2))) + (math.pow((5.1) - (input[2]), 2))) + (math.pow((1.5) - (input[3]), 2))))
    var23 = math.exp((var0) * ((((math.pow((5.8) - (input[0]), 2)) + (math.pow((2.8) - (input[1]), 2))) + (math.pow((5.1) - (input[2]), 2))) + (math.pow((2.4) - (input[3]), 2))))
    var24 = math.exp((var0) * ((((math.pow((6.1) - (input[0]), 2)) + (math.pow((3.0) - (input[1]), 2))) + (math.pow((4.9) - (input[2]), 2))) + (math.pow((1.8) - (input[3]), 2))))
    var25 = math.exp((var0) * ((((math.pow((7.7) - (input[0]), 2)) + (math.pow((2.6) - (input[1]), 2))) + (math.pow((6.9) - (input[2]), 2))) + (math.pow((2.3) - (input[3]), 2))))
    var26 = math.exp((var0) * ((((math.pow((6.9) - (input[0]), 2)) + (math.pow((3.1) - (input[1]), 2))) + (math.pow((5.1) - (input[2]), 2))) + (math.pow((2.3) - (input[3]), 2))))
    var27 = math.exp((var0) * ((((math.pow((6.3) - (input[0]), 2)) + (math.pow((3.3) - (input[1]), 2))) + (math.pow((6.0) - (input[2]), 2))) + (math.pow((2.5) - (input[3]), 2))))
    var28 = math.exp((var0) * ((((math.pow((4.9) - (input[0]), 2)) + (math.pow((2.5) - (input[1]), 2))) + (math.pow((4.5) - (input[2]), 2))) + (math.pow((1.7) - (input[3]), 2))))
    var29 = math.exp((var0) * ((((math.pow((6.0) - (input[0]), 2)) + (math.pow((2.2) - (input[1]), 2))) + (math.pow((5.0) - (input[2]), 2))) + (math.pow((1.5) - (input[3]), 2))))
    var30 = math.exp((var0) * ((((math.pow((7.9) - (input[0]), 2)) + (math.pow((3.8) - (input[1]), 2))) + (math.pow((6.4) - (input[2]), 2))) + (math.pow((2.0) - (input[3]), 2))))
    var31 = math.exp((var0) * ((((math.pow((7.2) - (input[0]), 2)) + (math.pow((3.0) - (input[1]), 2))) + (math.pow((5.8) - (input[2]), 2))) + (math.pow((1.6) - (input[3]), 2))))
    var32 = math.exp((var0) * ((((math.pow((7.7) - (input[0]), 2)) + (math.pow((3.8) - (input[1]), 2))) + (math.pow((6.7) - (input[2]), 2))) + (math.pow((2.2) - (input[3]), 2))))
    return [(((((((((((((((((((-0.08359187780790468) + ((var1) * (-0.0))) + ((var2) * (-0.0))) + ((var3) * (-0.4393498355605194))) + ((var4) * (-0.009465620856664334))) + ((var5) * (-0.16223369966927))) + ((var6) * (-0.26861888775075243))) + ((var7) * (-0.4393498355605194))) + ((var8) * (-0.4393498355605194))) + ((var9) * (-0.0))) + ((var10) * (-0.0))) + ((var11) * (-0.19673905328606292))) + ((var12) * (0.3340655283922188))) + ((var13) * (0.3435087305152051))) + ((var14) * (0.4393498355605194))) + ((var15) * (0.0))) + ((var16) * (0.28614124535416424))) + ((var17) * (0.11269159286168087))) + ((var18) * (0.0))) + ((var19) * (0.4393498355605194)), (((((((((((((((((((((-0.18563912331454907) + ((var20) * (-0.0))) + ((var21) * (-0.06014273244194299))) + ((var22) * (-0.0))) + ((var23) * (-0.031132453078851926))) + ((var24) * (-0.0))) + ((var25) * (-0.3893079321588921))) + ((var26) * (-0.06738007627290196))) + ((var27) * (-0.1225075748937126))) + ((var28) * (-0.3893079321588921))) + ((var29) * (-0.29402231709614085))) + ((var30) * (-0.3893079321588921))) + ((var31) * (-0.0))) + ((var32) * (-0.028242141062729226))) + ((var12) * (0.16634667752431267))) + ((var13) * (0.047772685163074764))) + ((var14) * (0.3893079321588921))) + ((var15) * (0.3893079321588921))) + ((var16) * (0.0))) + ((var17) * (0.0))) + ((var18) * (0.3893079321588921))) + ((var19) * (0.3893079321588921)), ((((((((((((((((((((((((0.5566649875797668) + ((var20) * (-25.563066587228416))) + ((var21) * (-38.35628154976547))) + ((var22) * (-38.35628154976547))) + ((var23) * (-0.0))) + ((var24) * (-38.35628154976547))) + ((var25) * (-0.0))) + ((var26) * (-0.0))) + ((var27) * (-0.0))) + ((var28) * (-6.2260303727828745))) + ((var29) * (-18.42781911624364))) + ((var30) * (-0.14775026537286423))) + ((var31) * (-7.169755983020096))) + ((var32) * (-0.0))) + ((var1) * (12.612328267927264))) + ((var2) * (6.565812506955159))) + ((var3) * (0.0))) + ((var4) * (38.35628154976547))) + ((var5) * (0.0))) + ((var6) * (38.35628154976547))) + ((var7) * (0.0))) + ((var8) * (0.0))) + ((var9) * (38.35628154976547))) + ((var10) * (38.35628154976547))) + ((var11) * (0.0))]
```

And here are some timings:
```
%%timeit -n 10000
score([1, 2, 3, 4])
```
```
310 µs ± 658 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

```
%%timeit -n 10000
score_raw([1, 2, 3, 4])
```
```
39.4 µs ± 136 ns per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Results seems to be identical:
```
np.testing.assert_allclose(score([1, 2, 3, 4]), score_raw([1, 2, 3, 4]))
```

Please share your thoughts about this refactoring.

StrikerRUS:
Closed #108.

See #108 for the rationale. With this PR `numpy` is used only for runtime code where some **vector math** is performed (that is for `numpy` was created).

I'm so sorry about the excess diff in `linear` and `xgboost` generated code examples for all languages. I guess they were caused by different package versions. Should I remove them from this PR or it's OK?

As a side note, I think there can be a CI job which will generate code examples with the latest packages and push them back into the repo after the each commit in `master`. So that examples will be always fresh.


m2cgen/interpreters/python/interpreter.py
```diff
@@ -9,9 +9,11 @@ class PythonInterpreter(ToCodeInterpreter,
     # 60 raises MemoryError for some SVM models with RBF kernel.
     bin_depth_threshold = 55
 
-    exponent_function_name = "np.exp"
-    power_function_name = "np.power"
-    tanh_function_name = "np.tanh"
+    exponent_function_name = "math.exp"
+    power_function_name = "math.pow"
+    tanh_function_name = "math.tanh"
```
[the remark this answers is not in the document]
krinart: Ohh.. I see, those are instances of `NumExpr` which means that they are scalars.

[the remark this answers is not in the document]
krinart: We should probably think about adding checks similar to what we have [here](https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/ast.py#L84-L85), but for subclasses of `NumExpr` to verify that they only accept/produce scalars.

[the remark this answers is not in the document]
StrikerRUS: > We should probably think about adding checks similar to what we have here, but for subclasses of NumExpr to verify that they only accept/produce scalars.

Yeah, seems to be so!
I guess all other languages rely on this implicit assumption now as well. For instance, these C functions work only with scalars:
https://github.com/BayesWitnesses/m2cgen/blob/30abeef339012c9e5ab9132513ebb9d6d6b7d0b4/m2cgen/interpreters/c/interpreter.py#L20-L22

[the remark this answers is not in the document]
StrikerRUS: Should I add asserts for scalars in this PR? Or it's better to have a separate one?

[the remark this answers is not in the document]
krinart: I would say it's up to you as it's not really relevant to or required by this PR. 

On the other hand I feel like it's just adding 4 asserts: 1 for `ExpExpr`, 1 for `TanhExpr` and 2 for `PowExpr`. 

So whatever works for you.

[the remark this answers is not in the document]
StrikerRUS: @krinart I added asserts in this PR, if you do not mind 🙂 .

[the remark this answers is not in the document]
krinart: Thanks for your effort! 👍

m2cgen/interpreters/python/interpreter.py
```diff
@@ -9,9 +9,11 @@ class PythonInterpreter(ToCodeInterpreter,
...
+
+    with_vector_operations = False
```
izeigerman: Not sure why can't we keep using `with_vectors` flag here that comes from `ToCodeInterpreter`?

m2cgen/interpreters/python/interpreter.py
```diff
@@ -27,19 +29,27 @@ def interpret(self, expr):
...
+        if self.with_vector_operations:
             self._cg.add_dependency("numpy", alias="np")
 
         return self._cg.code
 
     def interpret_bin_vector_expr(self, expr, **kwargs):
+        self.with_vector_operations = True
         return self._cg.infix_expression(
-            left=self._do_interpret(expr.left, **kwargs),
+            left=self._cg.array_convert_to_numpy(
```
izeigerman: I'm not sure I understand this change. Why can't keep using `vector_init` for this purpose?

m2cgen/interpreters/python/interpreter.py
```diff
@@ -27,19 +29,27 @@ def interpret(self, expr):
...
             op=expr.op.value,
-            right=self._do_interpret(expr.right, **kwargs))
+            right=self._cg.array_convert_to_numpy(
+                self._do_interpret(expr.right, **kwargs)))
 
     def interpret_bin_vector_num_expr(self, expr, **kwargs):
+        self.with_vector_operations = True
         return self._cg.infix_expression(
-            left=self._do_interpret(expr.left, **kwargs),
+            left=self._cg.array_convert_to_numpy(
```
izeigerman: Same as above

m2cgen/interpreters/python/interpreter.py
StrikerRUS: `with_vectors` means that some vectors are used. Refer to https://github.com/BayesWitnesses/m2cgen/blob/master/generated_code_examples/python/classification/linear.py as an example.
`with_vector_operations` means that some vector math happens. Refer to https://github.com/BayesWitnesses/m2cgen/blob/30abeef339012c9e5ab9132513ebb9d6d6b7d0b4/generated_code_examples/python/classification/random_forest.py#L29 as an example.

m2cgen/interpreters/python/interpreter.py
StrikerRUS: Because `[1, 2, 3] + [4, 5, 6] != [5, 7, 9]` in Python.

m2cgen/interpreters/python/interpreter.py
StrikerRUS: @izeigerman Oh, just realized that `with_vector_operations` is the same as `with_linaer_algebra`.
Updated in the latest commit.

StrikerRUS:
I reverted changes from generated examples that are not related to this PR. Also, I realized that `with_vector_operations` I introduced here is the same as `with_linear_algebra instead`, so I used it to determine when `numpy` is needed.

m2cgen/interpreters/python/interpreter.py
izeigerman: My question is rather why do we need raw Python lists in a generated code in the first place. Why can't it always be a numpy array?

izeigerman: I'm guessing that this the part where we may fallback to Python lists to eliminate the `numpy` dependency but I'd rather always have a consistent representation of the vector (especially considering that we may return it from the `score` function). So I'd argue that every time when we need vectors, it should be `numpy` arrays.

StrikerRUS: Yeah, this is about eliminating heavy redundant dependency from runtime. I think that this is quite more important rather than have consistent return value. There are many frameworks which return either `type1`, or `type2`, depending on some conditions. BTW, `score` at present may return `double` and `double[]`, so this can be treated as inconsistency as well.

I think that from the user side it's much easier to wrap `score` function call into `np.array()` only once, if they need consistent output, rather than manually edit generated function removing dozens of `np.assarray()` entries, if they do not have/want/able to install `numpy`.

StrikerRUS: *Joke* 😃 : It's time to change project's description: 
> Transform ML models into a native code (Java, C, Python, Go, JavaScript) with zero dependencies

izeigerman: > BTW, score at present may return double and double[], so this can be treated as inconsistency as well.
Well there is a distinction between scalar and vector, but here we're introducing different flavors of vectors. 

Additionally we break a backward compatibility here, which is not great. 

I'm positive about the benefit of using `math` vs `numpy` in certain scenarios, but I'm not fully convinced about the whole vector thing.

@krinart you're a better Python expert than I am. What's your opinion?

StrikerRUS: > Well there is a distinction between scalar and vector, but here we're introducing different flavors of vectors.
Additionally we break backward compatibility, which is not great.

Agree! I only want to say that the project has version `0.4.0`, which means that, according to the
semantic versioning, its API is subject to change.

izeigerman: After discussion with @krinart I think it may not be that bad in Python world. We also discussed the possibility of dropping numpy dependency altogether in favor of manually written vector multiplication like we do in other languages. This PR can be the first step in that direction. Let's merge it. @StrikerRUS thanks a lot for your effort 👍

StrikerRUS: @izeigerman  Thanks a lot for considering this change!
> We also discussed the possibility of dropping numpy dependency altogether in favor of manually written vector multiplication like we do in other languages. 

It's a great step towards isolated environments where it's not so straightforward to install any dependency! I guess they are the majority of the target audience of the package. As of now only vector addition and multiplication by number should be supported, it will not hurt the performance too much.

However, with the growth of supported algorithms number there can be needed more non-trivial vector operations where `numpy` speed will be crucial. So I think there can be an option which will provide users the possibility to choose either they need a runtime with zero dependencies but slow, or one with a few number of dependencies but significantly faster.

cc @krinart

PR #114, opened 2019-10-23
added support for PowerShell

StrikerRUS:
With this PR Windows users will be able to execute ML models from "command line" without the need to install any programming language (PowerShell is already installed in Windows).


README.md
```diff
@@ -25,11 +25,12 @@ pip install m2cgen
...
 ## Supported Languages
 
-- Python
-- Java
 - C
 - Go
+- Java
 - JavaScript
+- PowerShell
+- Python
```
StrikerRUS: Apply alphabetical order to help users easier find requested language.

izeigerman:
Wow, @StrikerRUS, thanks for this PR

.travis.yml
```diff
@@ -1,14 +1,29 @@
...
+  - 3.7
+
+env:
+  matrix:
+    - LANG=c
+    - LANG=go
+    - LANG=java
+    - LANG=javascript
+    - LANG=powershell
+    - LANG=python
```
StrikerRUS: Refer to https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts and the failed latest commit due to exceeding 50 minutes.

I guess with the growths of number of supported languages this change was inevitable. With the growths of number of supported algorithms there will be need to parallelize across them as well. However, it seems that parallelization across languages is enough for now.

krinart: Wow, this looks really cool!

One thing that bothers me is that we will run exactly same set of unit tests for all of these runs. While it might seem insignificant, I would still prefer for every run to be isolated.

Would it make sense to add another row here specifically to run unit tests? However we should keep in mind that only unit tests report code coverage (this is by design), and so only this single run would report it. I'm not sure about travis specifics and how it would affect the calculation of the result code coverage.

krinart: For some reason powershell takes much more time to run: <10 minutes against 40:
Screen Shot 2019-10-25 at 6 09 03 PM

StrikerRUS: > I'm not sure about travis specifics and how it would affect the calculation of the result code coverage.

Not sure, but I think that only the last one for PR/commit is taken into account.



Ah! Even with breakage in terms of languages one job hitted 50 minutes!


Hm, what do you think about moving to [Azure Pipelines](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/phases?view=azure-devops&tabs=yaml#timeouts)? I can create a minimal config here or in a separate PR. But you will have to sign up there.

krinart: > Not sure, but I think that only the last one for PR/commit is taken into account.

What I mean is that even for one PR/commit there are many builds, each potentially reporting different coverage. So if we separate unit and e2e tests, only subset of our builds will report coverage, and somehow they will be aggregated into single metric.

> what do you think about moving to Azure Pipelines?

I don't have a strong argument against, however I think it's worth checking what exactly makes powershell tests runnning so long as sometimes we need an ability to run e2e tests locally.

From pure technical point it could be one of the following:
- installing dependencies;
- code generation;
- execution.

Or some combination of those, of course :)

StrikerRUS: > So if we separate unit and e2e tests, only subset of our builds will report coverage, and somehow they will be aggregated into single metric.

You can click here and take a look. Seems that a coverage tool deals with this situation fine:

image


> I don't have a strong argument against …

Fine! Then I'd like to prepare a separate PR for this today.

> ... however I think it's worth checking what exactly makes powershell tests running so long as sometimes we need an ability to run e2e tests locally.

Installing `PowerShell Core` takes 13 seconds, so to be honest, I simply don't know, sorry 🙁 . I only can confirm that the same situation is in my local Windows environment with native PowerShell (not Core): awful timings.


image

StrikerRUS: > Then I'd like to prepare a separate PR for this today.

Can you please register at Azure Pipelines and push an empty `azure-pipelines.yml` file in `master`, so that it will be possible to see a result for my incoming PR? Otherwise, with that file in `master` builds will not be triggered.

StrikerRUS: @krinart Please see #116.

StrikerRUS:
Hi! Any plans for this PR?

Can the following be treated as a workaround?

```diff
diff --git a/.travis.yml b/.travis.yml
index a3d4633..aec110d 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,7 @@ env:
     - LANG=go
     - LANG=java
     - LANG=javascript
-    - LANG=powershell TASK=regression
-    - LANG=powershell TASK=classification
+    - LANG=powershell
     - LANG=python
     - LANG=visual_basic

@@ -52,4 +51,4 @@ script:
   - python setup.py install
   - rm -rfd m2cgen/
   - pytest -v tests/e2e/test_cli.py
-  - pytest -v tests/e2e/test_e2e.py -m $LANG -k "train_model_$TASK"
+  - pytest -v tests/e2e/test_e2e.py -m $LANG
diff --git a/tests/utils.py b/tests/utils.py
index 5e73679..37c805f 100644
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -145,6 +145,12 @@ def cartesian_e2e_params(executors_with_marks, models_with_trainers_with_marks,
         # to be clean.
         model = clone(model)

+        # PowerShell is extremely slow, so decrease test_fraction for it
+        if executor_mark.name == "powershell":
+            trainer_name = trainer.__name__
+            trainer = functools.partial(trainer, test_fraction=0.02)
+            trainer.__name__ = trainer_name
+
         # We use custom id since pytest for some reason can't show name of
         # the model in the automatic id. Which sucks.
         ids.append("{} - {} - {}".format(
```

Travis results for the patch above: https://travis-ci.org/StrikerRUS/m2cgen/builds/616146310 (all failures due to coverage token error).

izeigerman:
@StrikerRUS sorry, I didn't follow this PR much. @krinart can you please comment on that?

StrikerRUS:
@izeigerman FYI, here are two problems:
- PowerShell is **VERY** slow (~40 min vs 10 min in comparison with other languages);
- Travis has 50 min limit for 1 job (therefore, breakdown of tests in terms of languages/models is inevitable at some point in the future).

m2cgen/interpreters/powershell/code_generator.py
```diff
@@ -0,0 +1,70 @@
...
+    tpl_var_declare = CT("${var_type}${var_name} = ${init_val}")
+    tpl_var_assignment = CT("${var_name} = ${value}")
+    tpl_array_index_access = CT("$$${array_name}[${index}]")
+    tpl_return_statement = CT("return ${value}")
+
+    scalar_type = "[double]"
+    vector_type = "[double[]]"
+
+    operator_map = OrderedDict([("==", "-eq"), ("!=", "-ne"), (">=", "-ge"),
+                                ("<=", "-le"), (">", "-gt"), ("<", "-lt")])
```
StrikerRUS: Is it possible to have comparison operator templates as well to avoid such ugly hack like this one? For example, in different languages equality can be represented by `==`, `===`, `-eq`, inequality - by `!=`, `<>`, `-ne`, etc.

StrikerRUS: Added in the [latest commit](https://github.com/BayesWitnesses/m2cgen/pull/114/commits/14495ad54f571cab42fba7c2e01b2ed8d4d82126) possibility to overwrite default comparison operators via overriding `_comp_op_overwrite()` helper method of `CodeGenerator`.

Is it better now?

@izeigerman @krinart

StrikerRUS:
Resolved conflicts with `master` by old approach (separating into `regression` and `classification`). Still waiting for feedback on [this way to speed up test](https://github.com/BayesWitnesses/m2cgen/pull/114#issuecomment-557849702).

izeigerman:
Hey @StrikerRUS, with regard to the suggested workaround. I'm not sure I like the suggested solution that much.  
Shall we perhaps consider making Powershell a special case in E2E and configure it manually. 
In place of the the `<empty>` placeholder here: https://github.com/BayesWitnesses/m2cgen/blob/master/tests/e2e/test_e2e.py#L221 you can start listing your custom tests like this:
```
  pytest.param(	
        linear_regressor,	
        executors. PowershellExecutor,	
        utils.train_model_regression,	
        marks=[POWERSHELL, REGRESSION],	
    ),
    ...
```
This way you can generate a custom list of tests tailored specifically for Powershell. This will also allow us to tune tests for it better and perhaps give up some especially demanding tests altogether. 
What do you think?

StrikerRUS:
@izeigerman Thanks for your feedback!

We do not have any "special cases" for PowerShell. All tests are equally slow. Given that all tests are run row-by-row, I strongly believe that running all supported ML models on a smaller dataset is much more reliable unit testing than running a subset of supported models but on a bigger dataset.

izeigerman:
@StrikerRUS In this case I'd suggest to reduce the size of the test dataset for all tests and make them faster this way. Initially we didn't give much thought to the size of the test dataset since it didn't have much impact so we just didn't care as long as it was big enough. Perhaps it's a good time to revisit this value.

> but on a bigger dataset.
Not sure I completely agree here. Eg. in case of tree/ensemble models larger dataset ensures a higher likelihood of hitting more branches within each individual estimator.

StrikerRUS:
Yeah, there is always a tradeoff between time and quality/coverage!

> Perhaps it's a good time to revisit this value.

I think, in perfect, it should be easily adjustable value `test_fraction` which we can tune for different models/languages.

I'm happy to decrease `test_fraction` globally to push this PR forward, but I think that `0.02` (`10` / `3` / `11` samples) is quite small global value.

> Not sure I completely agree here. Eg. in case of tree/ensemble models larger dataset ensures a higher likelihood of hitting more branches within each individual estimator.

You are absolutely right! Larger datasets increases chances to cover more edge cases. However, I think that ensuring principal compatibility between a ML model and a language is more important than covering all code branches for a particular model.

izeigerman:
Ok, I think it's time to ship it. Thank you @StrikerRUS !

PR #119, opened 2019-11-15
added possibility to write generated code into file 

### LINKED ISSUE #110: add option to save generated code into file
I'm sorry if I missed this functionality, but `CLI` version hasn't it for sure (I saw the related code only in `generate_code_examples.py`). I guess it will be very useful to eliminate copy-paste phase, especially for large models.

Of course, piping is a solution, but not for development in Jupyter Notebook, for example.

StrikerRUS:
Closed #110.

Real-life frustrating example:
```
import sys

import lightgbm as lgb
import m2cgen as m2c

X, y = load_boston(True)
est = lgb.LGBMRegressor(n_estimators=1000).fit(X, y)

sys.setrecursionlimit(1<<30)
print(m2c.export_to_python(est))
```
```
IOPub data rate exceeded.
The notebook server will temporarily stop sending output
to the client in order to avoid crashing it.
To change this limit, set the config variable
`--NotebookApp.iopub_data_rate_limit`.

Current values:
NotebookApp.iopub_data_rate_limit=1000000.0 (bytes/sec)
NotebookApp.rate_limit_window=3.0 (secs)
```
`m2c.export_to_python(est, 'test.txt')` works fine in this scenario.


izeigerman:
Hey @StrikerRUS, thank you for your PR! 
I still don't completely get it though. Why complicate the API if it can be done by the user externally:
```
with open('test.txt', 'w') as fd:
    fd.write(m2c.export_to_python(est))
```

StrikerRUS:
Yeah, sure, it can be done so. But why not to do it for users?

And what about CLI interface? I don't think that it's common for programs to print huge output results to a user's screen, which you even cannot copy due to limited console history length. More often programs prefer to store result on disc. And I think that users are used to run programs in a such way.

StrikerRUS:
In other words, I want to say that you force users to perform one (or more) additional action in order to achieve their goal: piping, writing some code lines, copy-pasting and then saving... I can't imagine a case when user wants simply stare at a generated code 😃 .

m2cgen/exporters.py
```diff
@@ -19,92 +24,108 @@ def export_to_java(model, package_name=None, class_name="Model", indent=4):
...
         The model object that should be transpiled into code.
+    output_file : string, optional
+        Path to a file in which the generated code should be written.
     indent : int, optional
         The size of indents in the generated code.
 
     Returns
     -------
-    code : string
+    code : string or None
```
izeigerman: Let's document what's being returned when. Here and other places.

m2cgen/exporters.py
```diff
@@ -19,92 +24,108 @@ def export_to_java(model, package_name=None, class_name="Model", indent=4):
...
-def _export(model, interpreter):
+def _export(model, interpreter, output_file=None):
     assembler_cls = assemblers.get_assembler_cls(model)
     model_ast = assembler_cls(model).assemble()
-    return interpreter.interpret(model_ast)
+    code = interpreter.interpret(model_ast)
+    if output_file is None:
+        return code
+    else:
+        directory = os.path.dirname(output_file)
```
izeigerman: I'd say we should not only support string file names but file descriptors (`write()` method) as well.

izeigerman: I'd also add that I'm not sure that `m2cgen` should implicitly create directories.

krinart:
Hey @StrikerRUS 

Thanks again for this PR. I would like to share some thoughts and am interested in what you think.

[Unit philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) says:
1. Write programs that do one thing and do it well.
2. Write programs to work together.
3. Write programs to handle text streams, because that is a universal interface.

The reason to this is that there are millions of possible actions user would want to take after the code is generated. Saving code to the file is only one of them, although arguably the most common one.

With that being said I think the current implementation is exactly what Unix philosophy is about - we only do one thing, we do it well and we produce the text stream which can be used as an input to infinite number of other programs.

If we add the support for saving the result to the file, even such a seemingly simple use case has multiple options:
  - Saving to the file by file descriptor;
  - Saving to the file by filename. For this use case we would need to handle other edge cases:
    - What if file already exists
    - What if directory does not exist

Having this logic part of `m2cgen` will surely break the Unix philosophy principles, as now our responsibility is much broader than just producing an output for the other program's input.

The only way which is simple enough to incorporate into our codebase is saving results to the file by file descriptor. This way we don't need to handle any edge cases and we would simply write text into the file descriptor provided by user. 

For the API, if we decide to follow this path, I would prefer to have separate methods for saving to the file, i.e.:
  - `export_to_python_file`
  - `export_to_java_file`
 - etc.

The reason for this is imagine at some point we would like to extend the functionality of writing into file with extra param, without this approach we would either silently ignore this param or raise an exception that this param is not needed when exporting model into string. This would also prevent us from having functions returning different types in different circumstances.

StrikerRUS:
@krinart Thank you very much for your detailed opinion. Let me provide some comments.

>  I would like to share some thoughts and am interested in what you think.

I would say that it seems to me, programs with communication through stdin and stdout are either something old-fashioned from history, or something very specific from where nothing else is acceptable.


From my point of view, user experience is the most important thing here. So, if they save files with code very often, why can't we give this opportunity for them?

> If we add the support for saving the result to the file, even such a seemingly simple use case has multiple options:

I guess the most programs do that and we can see how they deal with it.

> The only way which is simple enough to incorporate into our codebase is saving results to the file by file descriptor.

I don't think that it worth to implement this. The key reason behind this PR is simplification of user's interaction with `m2cgen` and string file names seems to be the most user-friendly thing.

> For the API, if we decide to follow this path, I would prefer to have separate methods for saving to the file ..

I don't have anything against separate methods, but personally I prefer additional arguments. Simply because all arguments of one method are in the same place, while different methods depending on their names can pop up far away from each other in API reference or IDE code completion hints, which makes it hard to find all methods of one group.

> The reason for this is imagine at some point we would like to extend the functionality of writing into file with extra param, without this approach we would either silently ignore this param or raise an exception that this param is not needed when exporting model into string.

I guess you have already make a choice for the behavior in such situations when introduced in CLI language-specific arguments (`class_name`, `package_name`).

It is no doubt, that you guys have a better vision of `m2cgen` development direction. So, feel free to close this PR and the original issue with a link to the comment above as a reason explanation.

m2cgen/exporters.py
StrikerRUS: Sure, done!

krinart:
> I guess you have already make a choice for the behavior in such situations when introduced in CLI language-specific arguments (class_name, package_name).

I'm not sure I understand. My point is that if we mix exporting to string and to file, at some point we might introduce arguments that are specific to file export. Having language-specific arguments in language-specific methods seems to be the opposite of that approach.

> From my point of view, user experience is the most important thing here. So, if they save files with code very often, why can't we give this opportunity for them?

> I guess the most programs do that and we can see how they deal with it.

These are very valid reasons, let us think about them a bit more.

izeigerman:
```
I don't think that it worth to implement this. The key reason behind this PR is simplification of user's interaction with m2cgen and string file names seems to be the most user-friendly thing.
```
@StrikerRUS I'd say that the file name is too narrow of a use-case and if we're going to simplify things for a user anyway I'd say that supporting descriptors will be a natural extension of the reasoning behind this PR.

StrikerRUS:
> My point is that if we mix exporting to string and to file, at some point we might introduce arguments that are specific to file export. 

Sorry, seems that I didn't get it. Isn't it the same as using CLI version with following params `m2cgen --language=c --class_name=foo`? I mean, using specific param for one language with another one language.

> I'd say that the file name is too narrow of a use-case and if we're going to simplify things for a user anyway I'd say that supporting descriptors will be a natural extension of the reasoning behind this PR.

I agree that file descriptors can (should) be supported as well! I wonder how will they be used in CLI version?

izeigerman:
>  I wonder how will they be used in CLI version?
They won't be. They can be used via API however.

StrikerRUS:
> These are very valid reasons, let us think about them a bit more.

Hi guys! Any updates? I would like to know whether it's worth to continue working on this PR and what have you decided.

StrikerRUS:
Closing as stale.

PR #121, opened 2019-11-25
[VB] include Tanh function only when it is needed

StrikerRUS:
Do not include `Tanh` source code when only `Exp` is needed.


izeigerman:
Thanks 👍

PR #123, opened 2019-11-30
Limit the number of leaves in each subroutine for gradient boosted trees

izeigerman:


izeigerman:
Thanks for your contribution, @chris-smith-zocdoc 👍

PR #125, opened 2019-12-07
Split tests into different builds on Travis. Drop Python 3.4 support

izeigerman:


README.md
```diff
@@ -17,7 +17,7 @@
 * [FAQ](#faq)
 
 ## Installation
-Supported Python version is >= **3.4**.
+Supported Python version is >= **3.5**.
```
StrikerRUS: Two more places to update:
https://github.com/BayesWitnesses/m2cgen/blob/b8e4fac64e5f9267693c08294adc402b179802c2/setup.py#L28
https://github.com/BayesWitnesses/m2cgen/blob/b8e4fac64e5f9267693c08294adc402b179802c2/setup.py#L35

izeigerman: Good catch, thank you

StrikerRUS:
@izeigerman Thanks for fixing the problem of job time limit on Travis! Now I can contribute some more languages ^^

But don't you think that passing shell commands in string is not very convenient? It seems to me that it hurts readability significantly. Moreover, at Travis you cannot quickly understand what is failing anymore because env variables are extremely long and do not fit in UI.

image


I would like to suggest the following improvement:


```diff
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,18 +6,31 @@ python:
   - 3.6
   - 3.7

+env:
+  - TEST=API
+  - TEST=E2E LANG="c or python or java or go or javascript"
+  - TEST=E2E LANG="c_sharp or visual_basic"
+
 before_install:
-  - wget -q https://packages.microsoft.com/config/ubuntu/$(lsb_release -rs)/packages-microsoft-prod.deb
-  - sudo dpkg -i packages-microsoft-prod.deb
-  - sudo apt-get update
-  - sudo apt-get install --no-install-recommends -y dotnet-sdk-3.0
+  # Install .NET Core SDK.
+  - if [[ $LANG == "visual_basic" ]] || [[ $LANG == "c_sharp" ]]; then
+      wget -q https://packages.microsoft.com/config/ubuntu/$(lsb_release -rs)/packages-microsoft-prod.deb;
+      sudo dpkg -i packages-microsoft-prod.deb;
+      sudo apt-get update;
+      sudo apt-get install --no-install-recommends -y dotnet-sdk-3.0;
+    fi

 install:
   - pip install -r requirements-test.txt

-env:
-  - RUN_CMD="flake8 . && pytest -v tests/ --cov=m2cgen/ --ignore=tests/e2e/ && coveralls"
-  - RUN_CMD="python setup.py install && rm -rfd m2cgen/ && pytest -v -m 'c or python or java or go or javascript' tests/e2e/"
-  - RUN_CMD="python setup.py install && rm -rfd m2cgen/ && pytest -v -m 'c_sharp or visual_basic' tests/e2e/"
-
-script: "eval $RUN_CMD"
+script:
+  - if [[ $TEST == "API" ]]; then
+      flake8 .;
+      pytest -v tests/ --cov=m2cgen/ --ignore=tests/e2e/;
+      coveralls;
+    fi
+  - if [[ $TEST == "E2E" ]]; then
+      python setup.py install;
+      rm -rfd m2cgen/;
+      pytest -v -m $LANG tests/e2e/;
+    fi
```

It changes nothing in the logic but improves readability (IMHO).
In addition, it allows to save time during `before_install` by installing only the needed deps for the current language(s).

image

izeigerman:
Sure thing! Feel free to submit a PR

PR #126, opened 2019-12-11
control Travis workflow by env variables and do not store commands in them

StrikerRUS:
Refer to https://github.com/BayesWitnesses/m2cgen/pull/125#issuecomment-563389527.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,18 +6,31 @@ python:
   - 3.6
   - 3.7
 
+env:
+  - TEST=API
+  - TEST=E2E LANG="c or python or java or go or javascript"
+  - TEST=E2E LANG="c_sharp or visual_basic"
+
 before_install:
-  - wget -q https://packages.microsoft.com/config/ubuntu/$(lsb_release -rs)/packages-microsoft-prod.deb
-  - sudo dpkg -i packages-microsoft-prod.deb
-  - sudo apt-get update
-  - sudo apt-get install --no-install-recommends -y dotnet-sdk-3.0
+  # Install .NET Core SDK.
+  - if [[ $LANG == "c_sharp or visual_basic" ]]; then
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -6,18 +6,31 @@ python:
...
+  - TEST=E2E LANG="c or python or java or go or javascript"
+  - TEST=E2E LANG="c_sharp or visual_basic"
+
 before_install:
-  - wget -q https://packages.microsoft.com/config/ubuntu/$(lsb_release -rs)/packages-microsoft-prod.deb
-  - sudo dpkg -i packages-microsoft-prod.deb
-  - sudo apt-get update
-  - sudo apt-get install --no-install-recommends -y dotnet-sdk-3.0
+  # Install .NET Core SDK.
+  - if [[ $LANG == "c_sharp or visual_basic" ]]; then
```
izeigerman: I suggest we have a third flag like `IS_WIN_ENV=1/0` or similar for this purpose, so that when the new language comes in, it has to be added in a single place only.

izeigerman:
@StrikerRUS this is awesome, thanks a lot! One comment and let's ship it

.travis.yml
StrikerRUS: Sorry, I'm not sure I fully understood you. But in the latest commit I refined condition so that adding new language in string variable `LANG` will not require modifying condition. For example, adding `powershell` from #114 to `"c_sharp or visual_basic"` will not affect when DOT NET is installed. In other words, now it is possible to re-combine languages inside `LANG` variable to balance CI without the need to edit code on other places. I hope you mean that.

izeigerman: I believe the current solution works too. Thanks a lot 👍

StrikerRUS:
Thank you, @izeigerman !

PR #130, opened 2019-12-18
added LightGBM Random Forest mode

StrikerRUS:


m2cgen/assemblers/__init__.py
```diff
@@ -79,5 +83,8 @@ def get_assembler_cls(model):
     if not assembler_cls:
         raise NotImplementedError(
             "Model {} is not supported".format(model_name))
+    if assembler_cls is LightGBMModelAssembler:
+        if model.booster_.dump_model()["average_output"]:
```
izeigerman: Seems like a logic very specific to the `LightGBM` assembler. Do you think it can be handled in the `LightGBMModelAssembler`? Not sure whether it's worth creating  a separate `LightGBMRandomForestModelAssembler` entity.

StrikerRUS: Yeah, it will greatly simplify the logic! The only reason I created a separate class was that it's more logically correct to name it `ensemble` rather then `boosting`. But I think it's OK for the better maintain experience purposes. I just wan not sure that you'll agree with me.

PR #133, opened 2019-12-19
added XGBoost linear mode

StrikerRUS:


StrikerRUS:
I'm sorry, but I cannot fix code coverage decrease. The thing is that the same version of XGBoost in my Windows environment generates a different model (different thresholds), so I cannot write API test which will be valid at Travis with Linux environment.

I left a possibility to edit branch for repo maintainers, so feel free to push commits with tests if you feel like code coverage decrease is critical.

m2cgen/assemblers/boosting.py
```diff
@@ -194,6 +195,78 @@ def _count_leaves(tree):
         return num_leaves
 
 
+class XGBoostLinearModelAssembler(ModelAssembler):
```
izeigerman: A lot of this code looks a lot like `BaseBoostingAssembler`. Is there a way to generalize this? Let me know if you'd like to discuss this.

m2cgen/assemblers/__init__.py
```diff
@@ -83,4 +86,10 @@ def get_assembler_cls(model):
         raise NotImplementedError(
             "Model {} is not supported".format(model_name))
 
+    if assembler_cls is XGBoostModelAssembler:
+        model_dump = model.get_booster().get_dump(dump_format="json")
+        if len(model_dump) == 1 and all(i in json.loads(model_dump[0])
```
izeigerman: Looks like a logic that should reside within the assembler itself.

izeigerman:
Hey @StrikerRUS. Sorry about the delay. Holidays affected my ability to look into this.

m2cgen/assemblers/boosting.py
StrikerRUS: Yes, at the first glance it looks so. I started from inheriting `BaseBoostingAssembler` and failed to implement new functionality. So, I switched to `ModelAssembler`. The thing is that in `BaseBoostingAssembler` the main attribute and abstraction is `trees` which is passed everywhere. But in linear mode the main abstraction is `weight` and `bias`.

Please let me know if you have a proper way to implement this with `BaseBoostingAssembler` in your mind.

m2cgen/assemblers/__init__.py
StrikerRUS: Can you please elaborate a little bit more? Do you mean that the new class `XGBoostLinearModelAssembler` is not needed?

StrikerRUS:
@izeigerman 
> Hey @StrikerRUS. Sorry about the delay. Holidays affected my ability to look into this.

No problem! Thank you for looking into this! I left my comments above.

m2cgen/assemblers/__init__.py
izeigerman: I mean that we should have an umbrella `XGBoostModelAssembler` which will choose between the linear and tree version based on this logic. So I suggest 3 entities: `XGBoostTreeModelAssembler` - what is currently named `XGBoostModelAssembler`, `XGBoostLinearModelAssembler` - the implementation of which you've introduced and `XGBoostModelAssembler` - which contains the logic of choosing between `*Tree*` and `*Linear*`. 
What do you think?

m2cgen/assemblers/boosting.py
```diff
@@ -194,6 +195,78 @@ def _count_leaves(tree):
...
+            ast.NumVal(self._base_score),
+            expr,
+            ast.BinNumOpType.ADD)
+
+        return ast.SubroutineExpr(result_ast)
+
+    def _assemble_bin_class_output(self, weights, bias):
+        # Base score is calculated based on https://github.com/dmlc/xgboost/blob/master/src/objective/regression_loss.h#L64  # noqa
+        # return -logf(1.0f / base_score - 1.0f);
+        if self._base_score != 0:
```
izeigerman: Not a big fan of mutating a state during the operation which should be idempotent. Basically doing means that the same assembler instance can't be reused for different models, which is very far from obvious.

m2cgen/assemblers/boosting.py
izeigerman: I believe we'd have to introduce another entity on top of the `BaseBoostingAssembler`, eg. `BaseTreeBoostingAssembler` and modify the `BaseBoostingAssembler` in a way:
```
class BaseBoostingAssembler(ModelAssembler):
    ...
  def _assemble_estimators(self, *args, **kwargs):
    raise NotImplementedError

  def _assemble_single_output(self, base_score=0, *args, **kwargs):
    estimators_ast = self._assemble_estimators(*args, **kwargs)

    tmp_ast = utils.apply_op_to_expressions(
        ast.BinNumOpType.ADD,
        ast.NumVal(base_score),
        *estimators_ast)

    result_ast = self._final_transform(tmp_ast)

    return ast.SubroutineExpr(result_ast)
```
Same applies to `_assemble_bin_class_output` and so on.
Now for the `BaseTreeBoostingAssembler` the new method should look like following:
```
def _assemble_estimators(self, trees):
  if self._tree_limit:
    trees = trees[:self._tree_limit]

  trees_ast = [self._assemble_tree(t) for t in trees]
  
  # In a large tree we need to generate multiple subroutines to avoid
  # java limitations https://github.com/BayesWitnesses/m2cgen/issues/103.
  trees_num_leaves = [self._count_leaves(t) for t in trees]
  if sum(trees_num_leaves) > self._leaves_cutoff_threshold:
    return self._split_into_subroutines(trees_ast, trees_num_leaves)
  else:
    return trees_ast
```
And for the `XGBoostLinearModelAssembler`:
```
def _assemble_estimators(self, weights, bias):
  coef = utils.to_1d_array(weights)
  return [_linear_to_ast(coef, bias)]
```
Does this make sense? 
I can try to create a PR into your branch if that helps.

StrikerRUS: Yeah, sounds good!

> I can try to create a PR into your branch if that helps.

Please do it! It will help a lot and save much time, as code says more than words 🙂 .

m2cgen/assemblers/__init__.py
StrikerRUS: That sounds reasonable!

m2cgen/assemblers/boosting.py
izeigerman: Sorry about the delay. I'll try to come up with something over the upcoming weekend.

StrikerRUS: No worries! I don't think we are in a hurry. Please take as much time as you need!

StrikerRUS:
Superseded by #151.

PR #134, opened 2019-12-21
list models alphabetically

StrikerRUS:
It seems that with alphabetical listing it's easier to find a needed model.

rendered


izeigerman:
Good point, thank you 👍

PR #135, opened 2019-12-23
override comparison operators for VB

StrikerRUS:


StrikerRUS:
Closed in favor of #136.

PR #136, opened 2019-12-23
refined comparison operator overriding

StrikerRUS:


m2cgen/interpreters/code_generator.py
```diff
@@ -132,7 +132,7 @@ def _get_var_declare_type(self, expr):
         return NotImplemented
 
     def _comp_op_overwrite(self, op):
-        return op
+        return op.value
```
izeigerman: Considering that `_comp_op_overwrite` is only used in the interpreter and has little to do with the code generations, shall we perhaps consider moving it to the base interpreter class?

izeigerman: I gave it a second thought and realized that this has a direct relation to code generation since this is where we determine what operator to use, which is a part of a language syntax.

StrikerRUS: Huh, then I'm lucky that I had no time to refactor this yesterday!

PR #139, opened 2019-12-23
added support for R

StrikerRUS:
R is among of the most popular languages for data analysis.


izeigerman:
As always thank you for all the hard work 👍 
I somewhat struggle to come up with the use case when someone needs to generate R code from the trained model, considering that most of algorithm implementations (if not all of them) are supported by R one way or another. 
Do you perhaps keep in mind a particular use case when the code generation can be useful in R?

StrikerRUS:
I think that use cases here are the same as for Python: online-predictions which can be easily embedded into a web framework (e.g. https://github.com/microsoft/LightGBM/issues/2217); isolated environments where it's difficult to install third-party libraries; extremely visible and clean decision-making process. One more thing that is not applicable to Python, is that a lot of users have difficulties while installing LightGBM R-package as it is not at CRAN yet and requires manual compilation.

Despite all the above, I do not have any particular super useful use case for R right now. But I think when it will arise from one user, it will be too late. Because very often users search for their favorite feature (model in our case)/language support during working on a project which has time limitation, and cannot wait until someone pick up their feature request. So, they simply go away to another library 🙁 .

If you think that it's not worth to add R support, I'll abandon it without any worries and doubts, because I personally do not like this language 😄 .

izeigerman:
Thank you, @StrikerRUS, for the detailed clarification. I think what you've described makes sense. I wasn't aware of any issues that you enumerated due to a complete lack of experience with R and its ecosystem. All things considered, it seems like R may be a good addition to m2cgen.

StrikerRUS:
For the record: the blocking issue is `contextstack overflow` error
https://github.com/wch/r-source/blob/28fc79da885b3323e1f8ca6351736e15b96be551/src/main/gram.c#L3619

https://stackoverflow.com/a/25063770
> In this particular case, it is saying that you have more than 50 nested 
parse contexts -- that is nested if, [, [[, {, ( calls.
> [https://stat.ethz.ch/pipermail/r-help/2008-March/157341.html](https://stat.ethz.ch/pipermail/r-help/2008-March/157341.html)

StrikerRUS:
OK, inheriting `mixins.SubroutinesAsFunctionsMixin` solved the issue with random forest. However, inheriting `mixins.BinExpressionDepthTrackingMixin` cannot solve the `contextstack overflow` error in SVM, because it seems that **all** parentheses in expression are counted, not nested level:
```
from sklearn.svm import NuSVR

import m2cgen as m2c

X, y = load_boston(True)
est = NuSVR().fit(X, y)

print(m2c.export_to_r(est))
```
```
score <- function(input) {
    var4 <- (((((19.407718780443606) + ((subroutine0(input)) * (1.0))) + ((subroutine1(input)) * (1.0))) + ((subroutine2(input)) * (1.0))) + ((subroutine3(input)) * (-1.0))) + ((subroutine4(input)) * (-1.0))
    var5 <- (subroutine5(input)) * (-1.0)
    var3 <- ((((((((((((((((((((((((((((((((((((((((((((((((((var4) + (var5)) + ((subroutine6(input)) * (-1.0))) + ((subroutine7(input)) * (-1.0))) + ((subroutine8(input)) * (-1.0))) + ((subroutine9(input)) * (-1.0))) + ((subroutine10(input)) * (-1.0))) + ((subroutine11(input)) * (-1.0))) + ((subroutine12(input)) * (-1.0))) + ((subroutine13(input)) * (-1.0))) + ((subroutine14(input)) * (-1.0))) + ((subroutine15(input)) * (-1.0))) + ((subroutine16(input)) * (-1.0))) + ((subroutine17(input)) * (-1.0))) + ((subroutine18(input)) * (-1.0))) + ((subroutine19(input)) * (-1.0))) + ((subroutine20(input)) * (-1.0))) + ((subroutine21(input)) * (-1.0))) + ((subroutine22(input)) * (-1.0))) + ((subroutine23(input)) * (-1.0))) + ((subroutine24(input)) * (-1.0))) + ((subroutine25(input)) * (-1.0))) + ((subroutine26(input)) * (1.0))) + ((subroutine27(input)) * (1.0))) + ((subroutine28(input)) * (-1.0))) + ((subroutine29(input)) * (-1.0))) + ((subroutine30(input)) * (-1.0))) + ((subroutine31(input)) * (-1.0))) + ((subroutine32(input)) * (-1.0))) + ((subroutine33(input)) * (-1.0))) + ((subroutine34(input)) * (-1.0))) + ((subroutine35(input)) * (-1.0))) + ((subroutine36(input)) * (1.0))) + ((subroutine37(input)) * (1.0))) + ((subroutine38(input)) * (-1.0))) + ((subroutine39(input)) * (-1.0))) + ((subroutine40(input)) * (-1.0))) + ((subroutine41(input)) * (1.0))) + ((subroutine42(input)) * (-1.0))) + ((subroutine43(input)) * (-1.0))) + ((subroutine44(input)) * (1.0))) + ((subroutine45(input)) * (-1.0))) + ((subroutine46(input)) * (1.0))) + ((subroutine47(input)) * (1.0))) + ((subroutine48(input)) * (1.0))) + ((subroutine49(input)) * (1.0))) + ((subroutine50(input)) * (1.0))) + ((subroutine51(input)) * (1.0))) + ((subroutine52(input)) * (-1.0))) + ((subroutine53(input)) * (-1.0))) + ((subroutine54(input)) * (-1.0))
    var6 <- (subroutine55(input)) * (-1.0)
    var2 <- ((((((((((((((((((((((((((((((((((((((((((((((((((var3) + (var6)) + ((subroutine56(input)) * (-1.0))) + ((subroutine57(input)) * (-1.0))) + ((subroutine58(input)) * (-1.0))) + ((subroutine59(input)) * (-1.0))) + ((subroutine60(input)) * (-1.0))) + ((subroutine61(input)) * (-1.0))) + ((subroutine62(input)) * (-1.0))) + ((subroutine63(input)) * (-1.0))) + ((subroutine64(input)) * (-1.0))) + ((subroutine65(input)) * (-1.0))) + ((subroutine66(input)) * (-1.0))) + ((subroutine67(input)) * (-1.0))) + ((subroutine68(input)) * (-1.0))) + ((subroutine69(input)) * (-1.0))) + ((subroutine70(input)) * (1.0))) + ((subroutine71(input)) * (1.0))) + ((subroutine72(input))
...
```




I wonder, is there any way to limit the number of parentheses and braces in one expression right now?..

izeigerman:
@StrikerRUS nope, unfortunately there is no existing solution for parentheses :(

m2cgen/assemblers/boosting.py
```diff
@@ -42,7 +42,7 @@ def _assemble_single_output(self, trees, base_score=0):
         if self._tree_limit:
             trees = trees[:self._tree_limit]
 
-        trees_ast = [self._assemble_tree(t) for t in trees]
+        trees_ast = [ast.SubroutineExpr(self._assemble_tree(t)) for t in trees]
```
StrikerRUS: Is needed for both: to avoid `contextstack overflow` error and speeding up runtime.

m2cgen/interpreters/r/interpreter.py
```diff
@@ -0,0 +1,53 @@
...
+from m2cgen.interpreters.interpreter import ToCodeInterpreter
+from m2cgen.interpreters.r.code_generator import RCodeGenerator
+
+
+class RInterpreter(ToCodeInterpreter,
+                   mixins.LinearAlgebraMixin,
+                   mixins.BinExpressionDepthTrackingMixin,
+                   mixins.SubroutinesAsFunctionsMixin):
+
+    bin_depth_threshold = 25
```
StrikerRUS: Just a heuristic to pass tests with the hope that `50 / 2` will be enough for the most use cases.
Refer to https://github.com/BayesWitnesses/m2cgen/pull/139#issuecomment-569084447.
Asking users to re-compile R from sources is not a good idea I guess 😃 : 
> It is hardcoded (as 50) in 6 places in gram.c, so fairly easily changed. 
> https://r.789695.n4.nabble.com/contextstack-overflow-tp818869p818870.html

StrikerRUS:
Oops! I forgot to re-generate code examples for R. OK, will do it after a review to ensure that code reflects all updates.

m2cgen/assemblers/boosting.py
izeigerman: Sorry, what is `contextstack overflow`? Doesn't the `_split_into_subroutines` address the problem?

m2cgen/interpreters/r/code_generator.py
```diff
@@ -0,0 +1,36 @@
...
+        self.increase_indent()
+
+    @contextlib.contextmanager
+    def function_definition(self, name, args, is_vector_output):
+        self.add_function_def(name, args)
+        yield
+        self.add_block_termination()
+
+    def array_index_access(self, array_name, index):
+        return self.tpl_array_index_access(
```
izeigerman: Should this just be `return super(). array_index_access(array_name, index + 1)`?

m2cgen/interpreters/r/linear_algebra.r
```diff
@@ -0,0 +1,6 @@
+add_vectors <- function(v1, v2) {
```
izeigerman: I wonder why do we need this overhead together with the `LinearAlgebraMixin`, if we can just easily override `interpret_bin_vector_expr` and `interpret_bin_vector_num_expr` methods in place to provide these one-line expressions. Similarly to what we do in Python interpreter (with numpy). Am I missing something?

m2cgen/interpreters/r/interpreter.py
izeigerman: Yeah, if you can leave a comment with some reasoning behind this number - that'd be fantastic.

m2cgen/assemblers/boosting.py
StrikerRUS: For the description of `contextstack overflow` error please refer to https://github.com/BayesWitnesses/m2cgen/pull/139#issuecomment-568942961.

When I inherited `mixins.SubroutinesAsFunctionsMixin` (I guess it means that `_split_into_subroutines` applied automatically after that) it didn't solve the error, unfortunately. I wonder why each tree hadn't been wrapped into separate subroutine earlier?..

StrikerRUS:
@izeigerman Thank you very much for your detailed review! I addressed your comments in the latest commit.

m2cgen/interpreters/r/interpreter.py
```diff
@@ -0,0 +1,61 @@
+import os
+
+from m2cgen.interpreters import mixins, utils
+from m2cgen.interpreters.interpreter import ToCodeInterpreter
+from m2cgen.interpreters.r.code_generator import RCodeGenerator
+
+
+class RInterpreter(ToCodeInterpreter,
+                   mixins.LinearAlgebraMixin,
```
izeigerman: I believe this mixing is no longer needed, isn't it?

m2cgen/assemblers/boosting.py
izeigerman: Because it didn't address the problem, since an individual tree can be very large. We decided to rely purely on a `_split_into_subroutines` approach instead. Have you tried reducing the `leaves_cutoff_threshold` value here:
1. https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/assemblers/boosting.py#L13
2. https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/assemblers/boosting.py#L133
3. https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/assemblers/boosting.py#L201
(shoot, I should've had a named constant for this 🤦‍♂)

m2cgen/interpreters/r/interpreter.py
StrikerRUS: It's needed to set `self.with_linear_algebra = True`. Just like it's done in Python interpreter: https://github.com/BayesWitnesses/m2cgen/blob/9072ba010bbabfc8f9c9ad8bf9747f0c8a5d2af2/m2cgen/interpreters/python/interpreter.py#L40

m2cgen/assemblers/boosting.py
StrikerRUS: Ah, I see! Makes sense. 

Oops, seems I forgot about that threshold. Will try to reduce it shortly.

m2cgen/interpreters/r/interpreter.py
izeigerman: Ah, overlooked that. Sounds good 👍

izeigerman:
Btw, are you planning on adding generated samples in this PR or in a follow-up one?

m2cgen/assemblers/boosting.py
StrikerRUS: OK. I finally remembered why I ended up with decision of making "one tree - one subroutine". Refer to https://github.com/BayesWitnesses/m2cgen/pull/139#discussion_r361502439:
> Is needed for both to avoid `contextstack overflow` error and **speeding up runtime**.

Threshold 3000 is OK to avoid the error. But with this value Travis exceeds 10 minutes limit for no output. I just tried the following values of `leaves_cutoff_threshold`: [2000](https://travis-ci.org/StrikerRUS/m2cgen/jobs/636014647?utm_medium=notification&utm_source=github_status), [1500](https://travis-ci.org/StrikerRUS/m2cgen/jobs/636031082?utm_medium=notification&utm_source=github_status), [1000](https://travis-ci.org/StrikerRUS/m2cgen/jobs/636033835?utm_medium=notification&utm_source=github_status). Results are the same:
```
...

tests/e2e/test_e2e.py::test_e2e[LGBMRegressor - r_lang - train_model_regression_random_data] PASSED [ 28%]
tests/e2e/test_e2e.py::test_e2e[LGBMClassifier - r_lang - train_model_classification_random_data] 

No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

The build has been terminated
```

I think that checking smaller values makes no sense.

I would like to know WDYT.

StrikerRUS:
> Btw, are you planning on adding generated samples in this PR or in a follow-up one?

Any option which will be easier to review. FYI, `java` examples will be re-generated due to changes in boosting assembler.

m2cgen/assemblers/boosting.py
izeigerman: This is quite weird. I suggest we merge this now since the implementation works well otherwise and I'll try to investigate it further when I have time.

izeigerman:
@StrikerRUS I like the idea of introducing generated samples in a separate PR

m2cgen/assemblers/boosting.py
StrikerRUS: Sure!

I guess that slowness of `if` statement somehow is accumulated in a non-trivial way when a lot of `if` statements are placed inside one function. And splitting small trees into separate subroutines helps to speed up a runtime. I remember I saw one interesting topic which might be a good start of investigation when was googling fixes for `contextstack overflow`: https://stackoverflow.com/questions/34005423/is-if-faster-than-ifelse.

izeigerman:
I believe `setup.sh` is now lacking this block:
```
if [[ $LANG == *"r_lang"* ]]; then	
      sudo apt-get install --no-install-recommends -y r-base;	
fi
```

izeigerman:
Thanks for fixing conflicts 👍

PR #142, opened 2020-01-10
fix Travis to report failures

StrikerRUS:
Refer to https://github.com/travis-ci/travis-ci/issues/1066#issuecomment-32415710.

Seems that without this change Travis reflects only status of the last command in script in the job status.

See the current `master` build as an example of incorrectly reported status: https://travis-ci.org/BayesWitnesses/m2cgen/jobs/629771258?utm_medium=notification&utm_source=github_status.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,17 +14,21 @@ env:
 before_install:
   # Install .NET Core SDK.
   - if [[ $LANG == *"c_sharp"* ]] || [[ $LANG == *"visual_basic"* ]]; then
+      set -e;
       wget -q https://packages.microsoft.com/config/ubuntu/$(lsb_release -rs)/packages-microsoft-prod.deb;
       sudo dpkg -i packages-microsoft-prod.deb;
       sudo apt-get update;
       sudo apt-get install --no-install-recommends -y dotnet-sdk-3.0;
+      set +e;
     fi
   # Install PowerShell Core.
   - if [[ $LANG == *"powershell"* ]]; then
+      set -e;
       wget -q https://packages.microsoft.com/config/ubuntu/$(lsb_release -rs)/packages-microsoft-prod.deb;
       sudo dpkg -i packages-microsoft-prod.deb;
       sudo apt-get update;
       sudo apt-get install --no-install-recommends -y powershell;
+      set +e;
     fi
 
 install:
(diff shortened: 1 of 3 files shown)


izeigerman:
Oh my, this  is horrible. Thanks for catching this!

.travis.yml
```diff
@@ -14,30 +14,38 @@ env:
 before_install:
   # Install .NET Core SDK.
   - if [[ $LANG == *"c_sharp"* ]] || [[ $LANG == *"visual_basic"* ]]; then
+      set -e;
```
izeigerman: Shall we perhaps wrap only the beginning and the end of the `before_install` wrap?

.travis.yml
```diff
@@ -14,30 +14,38 @@ env:
...
       sudo apt-get install --no-install-recommends -y powershell;
+      set +e;
     fi
 
 install:
   - pip install -r requirements-test.txt
 
 script:
   - if [[ $TEST == "API" ]]; then
+      set -e;
```
izeigerman: Same as above: can we wrap the entire `script` block at once?

izeigerman:
Fixed unit tests in this PR https://github.com/BayesWitnesses/m2cgen/pull/143. Once that one is merged, this PR should become unblocked.

izeigerman:
I'd even say we only need `set -e` at the top of `before_install` and `script` blocks. We don't care about disabling errors back so `set +e` is somewhat redundant.

StrikerRUS:
I assume that `set +e` was suggested by Travis staff due to the following issue: https://github.com/travis-ci/travis-ci/issues/891. Seems that it has been successfully resolved and it's really safe to omit `set +e` now.

> I'd even say we only need `set -e` at the top of `before_install` and `script` blocks

Agree, if all steps inside blocks (they call them `phases`: https://docs.travis-ci.com/user/job-lifecycle/#customizing-the-build-phase) are run within the same bash session. I'll do experiments and report results here.

Or we can move commands in a separate `sh` files as it's suggested here: https://docs.travis-ci.com/user/job-lifecycle/#complex-build-commands.

izeigerman:
> Or we can move commands in a separate sh files as it's suggested here: https://docs.travis-ci.com/user/job-lifecycle/#complex-build-commands.
That might be a good idea. We could put those scripts into eg. `./travis` folder. 
Though right now I'm for whichever approach that will fix the build sooner.

StrikerRUS:
> That might be a good idea. We could put those scripts into eg. ./travis folder.

I'm for this approach too. I guess it'll be safer as one script file is guaranteed to be run within one shell session with less Travis-specific magic compared to config file's steps.

.travis/test.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+set -e
+
+cd $TRAVIS_BUILD_DIR
```
izeigerman: Just curious, why do we need this?

izeigerman:
Thanks for fixing this 🥇

.travis/test.sh
StrikerRUS: To not adjust all paths in commands, eg. `pytest -v ../tests/ --cov=../m2cgen/ --ignore=../tests/e2e/`.

PR #150, opened 2020-01-19
Fix support of older LightGBM versions

izeigerman:
Apparently the `average_output` attribute is not available in model dumps from older LightGBM versions (tested with 2.2.2).

### DIFF
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -208,10 +208,7 @@ class LightGBMModelAssembler(BaseBoostingAssembler):
         trees = [m["tree_structure"] for m in model_dump["tree_info"]]
 
         self.n_iter = len(trees) // model_dump["num_tree_per_iteration"]
-        if "average_output" in model_dump:
-            self.average_output = model_dump["average_output"]
-        else:
-            self.average_output = False
+        self.average_output = model_dump.get("average_output", False)
 
         super().__init__(model, trees,
                          leaves_cutoff_threshold=leaves_cutoff_threshold)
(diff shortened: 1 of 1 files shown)


StrikerRUS:
> Apparently the average_output attribute is not available in model dumps from older LightGBM versions

Confirm! That's why LightGBM's version was bumped in #130: https://github.com/BayesWitnesses/m2cgen/pull/130/files#diff-979cba9d30281710825398788016ac05R5.

Maybe we should indicate supported version ranges for supported (tested) models? It's cannot be done via `install_requires` in `setup.py` as we do not require installing them, but maybe somewhere in README?..

m2cgen/assemblers/boosting.py
```diff
@@ -206,8 +206,12 @@ def __init__(self, model,
                  leaves_cutoff_threshold=LEAVES_CUTOFF_THRESHOLD):
         model_dump = model.booster_.dump_model()
         trees = [m["tree_structure"] for m in model_dump["tree_info"]]
+
         self.n_iter = len(trees) // model_dump["num_tree_per_iteration"]
-        self.average_output = model_dump["average_output"]
+        if "average_output" in model_dump:
+            self.average_output = model_dump["average_output"]
+        else:
+            self.average_output = False
```
StrikerRUS:
```
self.average_output = model_dump.get("average_output", False)
```

izeigerman: Oh, yeah, that's a good point 👍

izeigerman:
Agree with the point about documenting versions

m2cgen/assemblers/boosting.py
izeigerman: Fixed

PR #155, opened 2020-01-23
remove scikit-learn and scipy from dependencies

StrikerRUS:
`scikit-learn` and `scipy` are quite "heavy" dependencies and rather often have conflicts with certain versions of other ML libraries. Dropping them from a list of required packages for `m2cgen` installation will allow users to be freer in configuring their virtual environments.

Some more my thoughts about the need of removing excess dependencies can be found in https://github.com/BayesWitnesses/m2cgen/issues/108#issue-508121116.

We do not require installation of `LightGBM` or `XGBoost`, for instance, to generate a code from their models. So I don't see any reasons why we should do this for `scikit-learn`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,7 @@ $ m2cgen <pickle_file> --language <language> [--indent <indent>] [--class_name <
          [--module_name <module_name>] [--package_name <package_name>] [--namespace <namespace>]
          [--recursion-limit <recursion_limit>]
 ```
+Don't forget that for unpickling serialized model objects their classes must be defined in the top level of an importable module in the unpickling environment.
 
 Piping is also supported:
 ```
@@ -118,3 +119,7 @@ $ cat <pickle_file> | m2cgen --language <language>
 **Q: Generation fails with `RuntimeError: maximum recursion depth exceeded` error.**
 
 A: If this error occurs while generating code using an ensemble model, try to reduce the number of trained estimators within that model. Alternatively you can increase the maximum recursion depth with `sys.setrecursionlimit(<new_depth>)`.
+
+**Q: Generation fails with `ImportError: No module named <module_name_here>` error while transpiling model from a serialized model object.**
+
+A: This error indicates that pickle protocol cannot deserialize model object. For unpickling serialized model objects, it is required that their classes must be defined in the top level of an importable module in the unpickling environment. So installation of package which provided model's class definition should solve the problem.
--- a/m2cgen/assemblers/tree.py
+++ b/m2cgen/assemblers/tree.py
@@ -3,8 +3,11 @@ import numpy as np
 from m2cgen import ast
 from m2cgen.assemblers import utils
 from m2cgen.assemblers.base import ModelAssembler
-from sklearn import tree
-from sklearn.tree._tree import TREE_LEAF
+
+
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -108,6 +108,7 @@ $ m2cgen <pickle_file> --language <language> [--indent <indent>] [--class_name <
          [--module_name <module_name>] [--package_name <package_name>] [--namespace <namespace>]
          [--recursion-limit <recursion_limit>]
 ```
+Don't forget that for unpickling serialized model objects their classes must be defined in the top level of an importable module in the unpickling environment.
```
StrikerRUS: Wordings are taken from the official documentation: https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled.

m2cgen/assemblers/tree.py
```diff
@@ -3,8 +3,11 @@
 from m2cgen import ast
 from m2cgen.assemblers import utils
 from m2cgen.assemblers.base import ModelAssembler
-from sklearn import tree
-from sklearn.tree._tree import TREE_LEAF
+
+
+# refer to 
+# https://github.com/scikit-learn/scikit-learn/blob/fd12d5684ad224ad7760374b1dcca2821c644feb/sklearn/tree/_tree.pyx#L64
+TREE_LEAF = -1
```
StrikerRUS: I'm absolutely sure that it doesn't worth to install huge library with own dependencies for one constant. 😃

m2cgen/assemblers/tree.py
```diff
@@ -13,7 +16,8 @@ def __init__(self, model):
         super().__init__(model)
         self._tree = model.tree_
         self._is_vector_output = False
-        if isinstance(self.model, tree.DecisionTreeClassifier):
+        if type(self.model).__name__ in {"DecisionTreeClassifier",
+                                         "ExtraTreeClassifier"}:
```
StrikerRUS: Now it's in consistency with a similar check in `boosting` assembler:
https://github.com/BayesWitnesses/m2cgen/blob/4c0f002885c65a2aac7725d64bfe47570b833248/m2cgen/assemblers/boosting.py#L25

m2cgen/assemblers/tree.py
izeigerman: Yes, I 100% agree with you.

README.md
```diff
@@ -118,3 +119,7 @@ $ cat <pickle_file> | m2cgen --language <language>
 **Q: Generation fails with `RuntimeError: maximum recursion depth exceeded` error.**
 
 A: If this error occurs while generating code using an ensemble model, try to reduce the number of trained estimators within that model. Alternatively you can increase the maximum recursion depth with `sys.setrecursionlimit(<new_depth>)`.
+
+**Q: Generation fails with `ImportError: No module named <module_name_here>` error while transpiling model from a serialized model object.**
+
+A: This error indicates that pickle protocol cannot deserialize model object. For unpickling serialized model objects, it is required that their classes must be defined in the top level of an importable module in the unpickling environment. So installation of package which provided model's class definition should solve the problem.
```
izeigerman: Wow, never ran into this one before. Thanks 👍

Issue #152, opened 2020-01-21
In Boosting Assembler wrapping each estimator into a subroutine causes a performance degradation


izeigerman:
I've recalled the real motivation behind not wrapping every individual estimator into its own subroutine - generation of many nested function calls leads to a performance degradation in Java. The observed difference reaches 4x for larger models (eg. XGBoost with 1000 estimators). The basic test I created (sorry about Scala):
```scala
@ import com.github.m2cgen.ModelOld
import com.github.m2cgen.ModelOld

@ import com.github.m2cgen.ModelNew
import com.github.m2cgen.ModelNew

@ def nextRandomData(): Array[Double] = (0 until 4).map(_ => Random.nextDouble).toArray
defined function nextRandomData

@ def testScore: Unit = {
    val start = System.currentTimeMillis()
    (0 until 100000).foreach(_ => <ModelNew|ModelOld>.score(nextRandomData))
    println("Runtime: " + (System.currentTimeMillis() - start).toString)
  }
```
Results for `ModelOld`:
```
@ testScore
Runtime: 2973
```
For `ModelNew`:
```
@ testScore
Runtime: 10747
```
The test model has been trained using the `sklearn.datasets.load_iris()` dataset. Classifier has been created as following:
```
model = XGBClassifier(n_estimators=1000)
```
In the attached [archive](https://github.com/BayesWitnesses/m2cgen/files/4092448/java_boosting_test.tar.gz) I included the following:
1. ModelNew.java - java code generated with the most recent master.
2. ModelOld.java - java code generated with the release `0.5.0` version.
3. Models.jar - the jar containing both compiled sources.
4. xgboost_model2 - the trained estimator in Pickle format.

CC: @StrikerRUS FYI

StrikerRUS:
x4 slowdown is very awful! However, Java slowdown due to wrapping estimators into subroutine should be less panful compared to R slowdown due to not wrapping, because slow Java code doesn't exceed Travis limit while R code do this.
Do you have any idea for possible compromise yet?

izeigerman:
Solving this issue on assembler's end rather than in interpreters certainly wasn't the best idea ever. So far I don't have any great ideas, but I'm sure I don't like penalizing one language in favor of the other.

izeigerman:
Turns out I've accidentally generated `ModelNew.java` for a wrong model. After rerunning my tests the numbers look a lot more reasonable for the current master:
```
@ testScore
Runtime: 3526
```
[...]

PR #158, opened 2020-01-26
In Java interpreter ignore subroutines and perform code split based on the AST size

izeigerman:
After investigating possible solutions for https://github.com/BayesWitnesses/m2cgen/issues/152, I came to a conclusion that with the existing design it's extremely hard to come up with the optimal algorithm to split code into subroutines on the interpreter side (and not in assemblers). 
The primary reason for that is that since we always interpret one expression at a time it's hard to predict both the depth of the current subtree and the number of expressions that are left to interpret in other branches.
I've achieved some progress by splitting expressions into separate subroutines based on the size of the code generated so far (i.e. code size threshold), but more often than not I'll get some stupid subroutines like this one:
```
public static double subroutine2(double[] input) {
    return 22.640634908349323;
}
```
That's why I took a simpler approach and attempted to optimize an interpreter that caused trouble in the first place - the R one. 
I slightly modified its behavior: when the binary expressions count threshold is exceeded, it no longer split them into separate variable assignments, but moves them into their own subroutines. Although it might not be the most optimal way for simpler models (like linear ones), it helps tremendously with gradient boosting and random forest models. 
Since those models are summation of independent estimators, we end up putting every N (5 by default) estimators into their own subroutine, improving this way the execution time.
@StrikerRUS please let me know what you think.


StrikerRUS:
@izeigerman  All your thoughts seems to be reasonable for me. However, R tests still have unacceptable execution time or even hang forever 🙁 .

izeigerman:
Yeah, apparently now it's being killed due to hitting the memory constraint :( I'm now seriously considering to limit the number of models tested for R.

izeigerman:
Ended up excluding large XGBoost and LGBM models from the R's test suite.

StrikerRUS:
OK, so removing subroutine wrapper for each tree actually means that XGBoost and LightGBM are not supported in R anymore - generated code for tiny model cannot be executed in a feasible time (I guess R session crashes, but Travis doesn't report anything for some reason and simply hangs forever). I don't think that it's a good alternative for slowness in Java.
I believe that it will be much better to continue support all models for both languages by adding a mechanism which will allow to filter languages. For instance, something like this: `ast.SubroutineExpr(self._assemble_tree(t), exclude_lang={'Java'})`. Or any other solution but not one which drops the support of subset of models.

izeigerman:
Hm
> generated code for tiny model cannot be executed in a feasible time (I guess R session crashes

I couldn't see any evidence to that.  Can you please share more context on this one?  What tiny models?

>  For instance, something like this: ast.SubroutineExpr(self._assemble_tree(t), exclude_lang={'Java'}).
This will introduce a tight coupling between assemblers and interpreters which is IMHO a poor design decision.

izeigerman:
Additionally it doesn't seem like R was designed for interpreting extensive amounts of code. We're pushing its limits already and I have a feeling it's not hard to come up with a large enough model that will blow up the R interpreter. 
So I suggest we admit that this platform is quite limited in its capabilities (either in memory constraints or runtime). Though I think it should still support smaller ensemble models. 
Can you please provide an example of the model which fails?

StrikerRUS:
> I couldn't see any evidence to that. Can you please share more context on this one?

I don't believe that [7.5GB](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system) is a harsh memory constraint you mentioned. Instead, I think it hangs due to some reason which is similar to `context overflow` and caused by a lot of `if`s in one function.

> What tiny models?

You excluded LightGBM with the following params `LIGHTGBM_PARAMS_LARGE = dict(n_estimators=100, num_leaves=100, max_depth=64, random_state=RANDOM_SEED)`. I don't think that they can be treated as a serious model in a real world application.

> This will introduce a tight coupling between assemblers and interpreters which is IMHO a poor design decision.

Agree! I just gave an example I imagined within 1m which will allow to save the support of LightGBM and XGBoost for R. Of course, I believe we will be able to develop quite smarter and more efficient solution.

izeigerman:
@StrikerRUS , somehow our conversation was very thought provoking. I've just got another idea and preliminary tests look very promising. Please ignore this PR for now and I'll tag you once it's ready to be revisited. Thanks!

StrikerRUS:
> Additionally it doesn't seem like R was designed for interpreting extensive amounts of code. We're pushing its limits already and I have a feeling it's not hard to come up with a large enough model that will blow up the R interpreter.

I think it's applicable to practically every programming language, especially interpreted one. 😃 

> So I suggest we admit that this platform is quite limited in its capabilities (either in memory constraints or runtime). Though I think it should still support smaller ensemble models.

Yes, but wrapping each tree into a function allows to support bigger models. Why should we consciously limit the number of supported models while we know how to overcome it? 
In addition, it's very naturally to do so. One tree - one function sounds super reasonable. For dealing with extremely deep trees we have a mixin which is useful addition.
It seems that Java has its problems too as it becomes slower with larger number of methods. And here we comes again to the problem about giving more love to one language...

> Can you please provide an example of the model which fails?

Sorry, didn't get it. Which model did you mean?

I strongly believe that we should continue wrapping trees into subroutines for R and don't do it for Java somehow. It looks like the best solution we could offer for users.

StrikerRUS:
Ah, I'm answering to an outdated comment again! 😄 Seems that something is wrong with my browser it doesn't provide new comments without pressing F5.

 >I've just got another idea and preliminary tests look very promising. 

Wow, it sounds very awesome! Will be happy to see it. Thank you very much for the constructive conversation!

izeigerman:
Hey @StrikerRUS, this PR is ready to be revisited.
It appears I've finally managed to come up with a good enough heuristics that is able to successfully compete with SubroutineExpr approach. I've compared models from E2E tests as well as some random configurations of boosting models and the number of generated Java methods is more or less the same. 
Of course the new approach is not as stable and predictable as the old one so outliers are to be expected.
Another downside of the new algorithm is that it's computationally demanding and it takes more time to interpret AST. At this point I consider this downside to be negligible.
In a follow-up PR I'm planning to do the following:
1. Generalize solution I came up with for Java and reuse it in R.
2. Drop the `SubroutineExpr` altogether.

Thanks!

m2cgen/interpreters/c/interpreter.py
```diff
@@ -15,6 +15,7 @@ class CInterpreter(ToCodeInterpreter,
 
     supported_bin_vector_num_ops = {
         ast.BinNumOpType.MUL: "mul_vector_number",
+        ast.BinNumOpType.DIV: "div_vector_number",
```
StrikerRUS: It seems to me that introducing new linear algebra function is not directly related to this PR but produces a huge diff which hurts reviewability a lot. Can you please move code changes connected to the new function into a separate PR?

m2cgen/assemblers/ensemble.py
```diff
@@ -7,17 +7,16 @@
...
+            return ast.SubroutineExpr(assembler.assemble())
 
         assembled_trees = [assemble_tree_expr(t) for t in trees]
-        return utils.apply_op_to_expressions(
-            ast.BinNumOpType.ADD, *assembled_trees)
+        return utils.apply_bin_op(
+            utils.apply_op_to_expressions(ast.BinNumOpType.ADD,
+                                          *assembled_trees),
+            ast.NumVal(self.model.n_estimators),
+            ast.BinNumOpType.DIV)
```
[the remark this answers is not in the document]
izeigerman: Yeah, totally. Overlooked it, thanks 👍

m2cgen/ast.py
```diff
@@ -230,3 +230,43 @@ def __init__(self, expr, to_reuse=False):
     def __str__(self):
         args = ",".join([str(self.expr), "to_reuse=" + str(self.to_reuse)])
         return "SubroutineExpr(" + args + ")"
+
+
+def count_exprs(expr, exclude_list=None):
```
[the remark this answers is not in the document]
izeigerman: I'm not sure I like the suggested approach but I see your point and have an alternative idea how to drop all the if's.

m2cgen/interpreters/c/interpreter.py
izeigerman: Although it's not too obvious, this change to ensemble model assembler is required in order for a new heuristics to perform at its best on models of this kind. This is why I changed the logic in the first place.
Unfortunately while iterating on this issue I did a terrible job at splitting logical changes into their corresponding commits. At this point coming up with a separate PR requires some time investment.
I 100% agree that I should've done a better job at splitting these 2, but it wasn't that obvious while I was working on the issue. Majority of the code changes is a copy-paste of `div` functions in tests. If it's not too much trouble for you to review, then I'm going to ask to keep this PR as is.

izeigerman: Otherwise the entire heuristics is only focused in the `java/interpreter.py` module + `ast.py` with expressions count logic.

izeigerman: Looking at LightGBM RF implementation I realized that I did a stupidest update ever :D I will revert div updates now, since we can achieve the same behavior with MUL. This is what happens when your focus is too narrow.

m2cgen/assemblers/ensemble.py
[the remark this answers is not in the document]
izeigerman: This actually does the right thing and I did a stupid unnecessary work. See https://github.com/BayesWitnesses/m2cgen/pull/158#discussion_r372172662

m2cgen/interpreters/c/interpreter.py
izeigerman: Should be much better now.

izeigerman: All I wanted to achieve was instead of doing this (a lot of expressions)
```
(1 / coef * a + 1 / coef * b + ... + 1 / coef * n)
```
to do this:
```
(a + b + ... + n) * (1 / coef)
```
But as a result of neglecting basic algebra and the distributive law in particular I ended up implementing this crap:
```
(a + b + ... + n) / coef
```

StrikerRUS: > This is what happens when your focus is too narrow.

Ahaha, I know what you mean! 😃

tests/test_ast.py
```diff
@@ -0,0 +1,40 @@
+from m2cgen import ast
+
+
+def test_count_exprs():
```
StrikerRUS: Can you please add one really big expression which contains all expressions we have one or more times?

.travis.yml
```diff
@@ -8,8 +8,9 @@ python:
 
 env:
   - TEST=API
-  - TEST=E2E LANG="c or python or java or go or javascript or r_lang"
-  - TEST=E2E LANG="c_sharp or visual_basic or powershell or php"
+  - TEST=E2E LANG="c or python or java or go or javascript or php"
+  - TEST=E2E LANG="c_sharp or visual_basic or powershell"
+  - TEST=E2E LANG="r_lang"
```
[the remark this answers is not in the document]
izeigerman: Hm, looks balanced enough to me:
Screen Shot 2020-01-30 at 7 51 52 AM

m2cgen/ast.py
```diff
@@ -230,3 +230,33 @@ def __init__(self, expr, to_reuse=False):
     def __str__(self):
         args = ",".join([str(self.expr), "to_reuse=" + str(self.to_reuse)])
         return "SubroutineExpr(" + args + ")"
+
+
+NESTED_EXPRS_MAPPINGS = [
```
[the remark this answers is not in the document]
izeigerman: Few things:
1. The existing implementation of the algorithm is self-contained and is not spread across dozen classes.
2. I'd really love to keep existing AST entities concise and not a pollute a data structure definition with random behavior. Moving the count logic inside will create a precedent and future temptation to throw more stuff in there.
3. With `exclude_list` parameter each entity will have to check whether it's not excluded, or otherwise check its branches. This logic will have to reside in a base class or something, which again introduces extra complexity.

.travis.yml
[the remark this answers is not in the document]
izeigerman: Or am I missing a point?

m2cgen/ast.py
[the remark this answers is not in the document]
izeigerman: 4. I'm coming from functional programming background and seeing AST as [ADTs](https://en.wikipedia.org/wiki/Algebraic_data_type), which I just realized may not be as appropriate in Python world :D

.travis.yml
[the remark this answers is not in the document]
StrikerRUS: Oops, sorry! I was holding previous numbers before refactoring in my mind. New values seem to be really balanced.

m2cgen/ast.py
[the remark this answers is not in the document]
StrikerRUS: Many thanks for the detailed explanation! It was really interesting to hear your vision.

izeigerman:
@StrikerRUS Sorry about the delay, this should be good to go. Thanks!

m2cgen/ast.py
```diff
@@ -230,3 +230,34 @@ def __init__(self, expr, to_reuse=False):
...
+    ((BinExpr, CompExpr), lambda e: [e.left, e.right]),
+    (PowExpr, lambda e: [e.base_expr, e.exp_expr]),
+    (VectorVal, lambda e: e.exprs),
+    (IfExpr, lambda e: [e.test, e.body, e.orelse]),
```
StrikerRUS: I think these expr can be wrapped into a tuples too for the consistent interface. Like, `((PowExpr), lambda e: [e.base_expr, e.exp_expr]),`

tests/test_ast.py
```diff
@@ -0,0 +1,70 @@
...
+    ) == 0
+
+    assert ast.count_exprs(
+        ast.BinNumExpr(ast.NumVal(1), ast.NumVal(2), ast.BinNumOpType.ADD),
+        exclude_list={ast.BinNumExpr}
+    ) == 2
+
+
+def test_count_all_exprs_types():
+    expr = ast.BinVectorNumExpr(
```
StrikerRUS: Maybe even more complicated expr with deeper nesting? 🙂

izeigerman: TBH, I don't see a value in having even deeper nesting since it won't really provide additional coverage. Additionally this will make the test more complex and somewhat obscure its purpose. Thanks for the feedback though.

m2cgen/ast.py
izeigerman: Sorry, I'm not 100% sold on this :D

izeigerman:
Thanks for your review 👍

PR #159, opened 2020-01-28
added support for base linear regression models from statsmodels

StrikerRUS:
[StatsModels](https://github.com/statsmodels/statsmodels) is one of the most popular Python package for statistics. And from my personal point of view, it is the most popular one for **pure** statistics (or in other words, among pure statisticians 🙂 ). A brief comparison (rather outdated, but still ...) of scikit-learn and StatsModels: https://blog.thedataincubator.com/2017/11/scikit-learn-vs-statsmodels/.

This PR adds the basic support for [linear regression models](https://www.statsmodels.org/stable/api.html#regression) from statsmodels package: OLS, WLS and GLS. Unfortunately, my experience in this package is quite limited - I used it only several times. That's why this PR includes the support for only 3 base models. Later I have plans to verify code generation for the rest models which produce [`RegressionResultsWrapper`](https://github.com/statsmodels/statsmodels/blob/93753a3e750c2530df2efff41bf488fc26f3fe4f/statsmodels/regression/linear_model.py#L1471) and [`RegularizedResultsWrapper`](https://github.com/statsmodels/statsmodels/blob/1f8434af5d9ee5707ded7ecbc746de93801fe8af/statsmodels/base/elastic_net.py#L375) and add more models.

Refactoring of training routines for tests allows to
- save time during datasets construction/loading phase as they are loaded only once;
- access training/test data shapes required to calculate models' initial params (the main reason to rewrite code).


StrikerRUS:
I think it's ready for review now.

tests/utils.py
```diff
@@ -21,6 +22,102 @@
...
+        self.fit_intercept_ = init_params.pop("fit_intercept", False)
+        if self.fit_intercept_:
+            X = sm.add_constant(X)
+        est = self.model(y, X, **init_params)
+        if "fit_regularized" in self.params:
+            self.fitted_model_ = est.fit_regularized(
+                **self.params["fit_regularized"])
+        else:
+            self.fitted_model_ = est.fit(**self.params.get("fit", {}))
+        self.__class__.__name__ = type(self.fitted_model_).__name__
```
izeigerman: This is something I don't fully follow. I assume that the type of the fitted model is different from the one passed into the constructor. I think I get it when we do this in constructor - so that `get_assembler_cls` can pick the right assembler, but why do we do this here? 

**UPD** I guess this is needed to make a model name in E2E tests look nicer (eg. OLS, GLS, etc.). Please correct me if I'm wrong.

tests/utils.py
```diff
@@ -21,6 +22,102 @@
...
+class ModelTrainer:
+
+    _class_instances = {}
+
+    def __init__(self, dataset_name, test_fraction):
+        self.dataset_name = dataset_name
+        self.test_fraction = test_fraction
+        np.random.seed(seed=7)
+        if dataset_name == "boston":
+            self.__name__ = "train_model_regression"
```
izeigerman: Seems more like a workaround to me. Can we make it an actual `@property` like `dataset_name` instead? I assume it's only used in a single place here: https://github.com/BayesWitnesses/m2cgen/blob/master/tests/utils.py#L191. We just need to use `trainer.dataset_name` instead of the `trainer.__name__`

tests/utils.py
```diff
@@ -21,6 +22,102 @@
...
+            self.y = np.random.randint(2, size=(N,))
+        else:
+            raise ValueError("Unknown dataset name: {}".format(dataset_name))
+
+        offset = int(self.X.shape[0] * (1 - test_fraction))
+        self.X_train, self.y_train = self.X[:offset], self.y[:offset]
+        self.X_test, self.y_test = self.X[offset:], self.y[offset:]
+
+    @classmethod
+    def get_instance(cls, dataset_name, test_fraction=0.02):
```
izeigerman: I like this caching mechanism 👍

tests/utils.py
izeigerman: Perhaps worth leaving an explanatory comment in both places, so that we don't have same questions 6 months from now :D

tests/utils.py
```diff
@@ -21,6 +22,102 @@
...
+        key = dataset_name + " {}".format(test_fraction)
```
izeigerman: This key bothers me a little. Shouldn't it be unique per dataset, fraction and a model? The model part seems to be missing here.

izeigerman: Oops, sorry, taking this back. Model is passed later. Please ignore.

tests/utils.py
izeigerman: Shall we consider making the fitted model a part of the cache as well? Considering that all models are shared between languages. Let me know what you think

tests/utils.py
StrikerRUS: > UPD I guess this is needed to make a model name in E2E tests look nicer (eg. OLS, GLS, etc.). Please correct me if I'm wrong.

You are absolutely right!

>Perhaps worth leaving an explanatory comment in both places, so that we don't have same questions 6 months from now :D

OK!

tests/utils.py
StrikerRUS: Sure! Good catch! 👍

tests/utils.py
StrikerRUS: Thanks!

Seems reasonable to me! I'll try.

StrikerRUS: I've pushed an attempt to cache fitted models in 27ffdd6a588bd0142dd918b71c51d8188a420454. All e2e tests are passed, however, some API tests are failing due to the usage of different params for the same class. The thing is that now I cache **class name** and therefore we cannot differentiate between class instances. Caching **class instances** seems to be pointless as we never use the same instance more than once. Please let me know if you have any workarounds in the mind.

StrikerRUS: To be honest, I'm starting to doubt that caching models is a good idea...

izeigerman: > To be honest, I'm starting to doubt that caching models is a good idea...

That's ok, it might as well be not a good idea indeed :)

> The thing is that now I cache class name and therefore we cannot differentiate between class instances. 

Can't you use the estimator's object pointer or some other instance identifier (eg. using `id()`) instead? I thought the point was to prevent fitting the same models more than once in E2E tests. Somehow I thought this approach should help. Am I missing something here?

izeigerman: I'd say it's a stretch goal anyway. We can revisit this part later. Shouldn't be a stopper for this PR.

StrikerRUS: > Can't you use the estimator's object pointer or some other instance identifier (eg. using id()) instead? I thought the point was to prevent fitting the same models more than once in E2E tests. Somehow I thought this approach should help. Am I missing something here?

I'm sorry, but I can't get the point of using `id()` (or some unique identifier) as a part of a caching key. Every time we create new estimator object in tests 
https://github.com/BayesWitnesses/m2cgen/blob/d28cb0063ab047751bd72eb7788b109a629f9068/tests/e2e/test_e2e.py#L131
we get new unique key, which results in a cache miss.
```
import lightgbm as lgb
a = lgb.LGBMRegressor()
b = lgb.LGBMRegressor()
id(a) == id(b)
>>> False
```

I tried the following way:
```
        key = dataset_key + " {}".format((hash(type(estimator).__name__
                                          + estimator.__dict__.__repr__()))
                                         if estimator is not None
                                         else "None")
```
But it didn't work for some API tests, because some classes have identical interfaces.

https://travis-ci.org/StrikerRUS/m2cgen/builds/650970719
image

Below we can see that for `SVR` model test it uses cashed previously trainer for `NuSVR`.
image


If you don't mind, I'm going to revert model caching for now as it's bugging and I'm out of ideas how to fix it. We can get back to it later.

izeigerman: > If you don't mind, I'm going to revert model caching for now as it's bugging and I'm out of ideas how to fix it. We can get back to it later.

Yes absolutely. Thank you so much for putting effort into this

StrikerRUS:
Thank you for giving it one more round of review!

PR #162, opened 2020-01-30
added support for RANSACRegressor

StrikerRUS:


m2cgen/assemblers/linear.py
```diff
@@ -22,6 +22,15 @@ def _build_ast(self):
         return ast.VectorVal(exprs)
 
 
+class RANSACModelAssembler(ModelAssembler):
```
StrikerRUS: Not sure, may it's better to place this class into a separate module, let say `meta.py`, as it has nothing in common with linear models implementation...

izeigerman: Perhaps in this particular case it makes sense to unwrap the estimator right inside of the `get_assembler_cls` function, and don't introduce any extra entities on top of it. 
My reasoning here is based on the fact that in this case no additional logic is introduced by RANSAC specifically, only unwrapping takes place.
What do you think?

izeigerman:
Thanks @StrikerRUS ! Sorry about the delay with PRs, was traveling these past 2 weeks. Left one comment, otherwise good to go.

m2cgen/assemblers/linear.py
StrikerRUS: I tried the approach you are talking about firstly 🙂  Unfortunately, it failed due to that we can get assembler for the underlying model in `get_assembler_cls()`, but we cannot replace `model` with `model.estimator_` there. So, we we'll try to assemble `RANSACRegressor` (still) instance with the appropriate assembler (`LinearModelAssembler` let say for example).
https://github.com/BayesWitnesses/m2cgen/blob/1553c4cd0fb2493720e70639d460927110e7fc90/m2cgen/exporters.py#L255-L256

Also, I think that `RANSACRegressor` is not the last meta estimator we might want to support. The first thing that comes to my mind is `GridSearchCV`. We can add the support for it right now simply by changing `base_model = self.model.estimator_` to `base_model = self.model.best_estimator_`. That's why I want to move the implementation to `meta.py` and make generalizable approach for meta estimators.

izeigerman: > but we cannot replace model with model.estimator_ there.
Ah, good point indeed. 
I think your reasoning makes a lot of sense. I also like the idea of moving it to `meta.py` module 👍

StrikerRUS: Thanks! Done in the latest commit.

PR #166, opened 2020-02-14
add: Make function_name parametrized

mrshu:
Hello everyone,

First of all, thanks a ton for putting this tool/library together -- especially in resource-stranded environments, it does have a potential to literally save lives!

One small problem I was fighting with while using it was the `score` function it uses in the generated modules. When they are used as drop-in replacements for trained models, using `score` is a bit strange, as the API generally provides function like `predict` or `predict_proba`. It would therefore be of great help to me if this name could be dynamically changed and I would not have to do so manually.

Please do let me know if something like this sounds like a sensible addition. I'd be happy to update the code so that it reflect your vision, so please feel free to just let me know whenever that may be the case.

Thanks!

-------------------------------------------------------

* Currently `m2cgen` generates a module in various languages that has a
  "score"/"Score" function/method. This is not always desirable, as many
  of the trained modules that are to be exported may provide their
  prediction via API functions with different names (such as `predict`).

* This commit adds a way of specifying the name of the function both via
  the CLI and in the exporters (that is, in the `export_to_` funcitons)
  by specifying the `function_name` option/parameter while keeping the
  default set to "score"/"Score" for backwards compatilibity.

Signed-off-by: mr.Shu <mr@shu.io>

### DIFF
--- a/m2cgen/cli.py
+++ b/m2cgen/cli.py
@@ -52,6 +52,9 @@ parser.add_argument(
     required=True)
 parser.add_argument(
     "--function_name", "-fn", dest="function_name", type=str,
+    # The default value is conditional and will be set in the argument's
+    # post-processing, based on the signature of the `export` function
+    # that belongs to the specified target language.
     default=None,
     help="Name of the function in the generated code.")
 parser.add_argument(
--- a/m2cgen/cli.py
+++ b/m2cgen/cli.py
@@ -39,39 +39,39 @@ MAX_RECURSION_DEPTH = np.iinfo(np.intc).max
 
 parser = argparse.ArgumentParser(
     prog="m2cgen",
-    description="Generate code in native language for provided model")
+    description="Generate code in native language for provided model.")
 parser.add_argument(
     "infile", type=argparse.FileType("rb"), nargs="?",
     default=sys.stdin.buffer,
-    help="File with pickle representation of the model")
+    help="File with pickle representation of the model.")
 parser.add_argument(
     "--language", "-l", type=str,
     choices=LANGUAGE_TO_EXPORTER.keys(),
-    help="Target language",
+    help="Target language.",
     required=True)
 parser.add_argument(
     "--function_name", "-fn", dest="function_name", type=str,
     default="score",
-    help="Name of the function in the generated code (defaults to 'score')")
+    help="Name of the function in the generated code.")
 parser.add_argument(
     "--class_name", "-cn", dest="class_name", type=str,
-    help="Name of the generated class (if supported by target language)")
+    help="Name of the generated class (if supported by target language).")
 parser.add_argument(
     "--package_name", "-pn", dest="package_name", type=str,
     help="Package name for the generated code "
-         "(if supported by target language)")
+         "(if supported by target language).")
 parser.add_argument(
     "--module_name", "-mn", dest="module_name", type=str,
     help="Module name for the generated code "
-         "(if supported by target language)")
+         "(if supported by target language).")
 parser.add_argument(
     "--namespace", "-ns", dest="namespace", type=str,
     help="Namespace for the generated code "
-         "(if supported by target language)")
+         "(if supported by target language).")
 parser.add_argument(
     "--indent", "-i", dest="indent", type=int,
     default=4,
-    help="Indentation for the generated code")
+    help="Indentation for the generated code.")
 parser.add_argument(
     "--recursion-limit", "-rl", type=int,
     help="Sets the maximum depth of the Python interpreter stack. "
--- a/m2cgen/cli.py
+++ b/m2cgen/cli.py
@@ -11,6 +11,7 @@ Model can also be piped:
 import pickle
 import argparse
 import sys
+import inspect
 import numpy as np
 
 import m2cgen
@@ -51,7 +52,7 @@ parser.add_argument(
     required=True)
 parser.add_argument(
     "--function_name", "-fn", dest="function_name", type=str,
-    default="score",
+    default=None,
     help="Name of the function in the generated code.")
 parser.add_argument(
     "--class_name", "-cn", dest="class_name", type=str,
@@ -99,6 +100,16 @@ def generate_code(args):
         arg_value = getattr(args, arg_name)
         if arg_value is not None:
             kwargs[arg_name] = arg_value
+
+        # Special handling for the function_name parameter, which needs to be
+        # the same as the default value of the keyword argument of the exporter
+        # (this is due to languages like C# which prefer their method names to
+        # follow PascalCase unlike all the other supported languages -- see
+        # https://github.com/BayesWitnesses/m2cgen/pull/166/files#r379867601
+        # for more).
+        if arg_name == 'function_name' and arg_value is None:
+            param = inspect.signature(exporter).parameters['function_name']
+            kwargs[arg_name] = param.default
 
     return exporter(model, **kwargs)
 
--- a/m2cgen/cli.py
+++ b/m2cgen/cli.py
@@ -17,18 +17,19 @@ import m2cgen
 
 
 LANGUAGE_TO_EXPORTER = {
-    "python": (m2cgen.export_to_python, ["indent"]),
-    "java": (m2cgen.export_to_java, ["indent", "class_name", "package_name"]),
-    "c": (m2cgen.export_to_c, ["indent"]),
-    "go": (m2cgen.export_to_go, ["indent"]),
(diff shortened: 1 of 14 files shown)


izeigerman:
@mrshu Thanks a lot for submitting this PR together with such a detailed description!
I think this is a valuable updated and as far as I remember @StrikerRUS mentioned the importance of it some time ago too.
I'll take at it soon.

mrshu:
Thank you for your kind words @izeigerman!

Right off the bat I'd like to tell you that the way the `--function_name` CLI option is handed probably introduces a bug -- it currently always defaults to "score" which was not the case for C# and Powershell before. To ensure that it works as before we'd have to deal with some conditional defaults (as in, after `--language` is set and `--function_name` is set, ensure `Score` is the default in case `--language` was `C#` or `Powershell`) and I was not exactly sure where would something like that make the most sense.

I am therefore looking forward to your review and any further discussion!

mrshu:
Also, I am not completely sold on calling it `function_name` -- [`sklearn-porter`](https://github.com/nok/sklearn-porter#cli) for instance uses `--method_name` which may make a bit more sense in this case as well. Please do let me know what you think and I'll be happy to switch it over.

Thanks!

StrikerRUS:
@izeigerman 

> I think this is a valuable updated and as far as I remember @StrikerRUS mentioned the importance of it some time ago too.

To be honest, I don't remember I was saying this 😃 
I remember I spoke about **argument** name: https://github.com/BayesWitnesses/m2cgen/pull/109#discussion_r336280288.

@mrshu 
>To ensure that it works as before we'd have to deal with some conditional defaults 

I'm not sure that it is worthy of further consideration. As we are still in alpha phase, we are free to change a lot without any deprecation warnings and worrying about the backward compatibility. Especially for just a naming conventions.
https://github.com/BayesWitnesses/m2cgen/blob/d28cb0063ab047751bd72eb7788b109a629f9068/setup.py#L20

Please add a test for your PR. You may use the following tests as examples:
https://github.com/BayesWitnesses/m2cgen/blob/d28cb0063ab047751bd72eb7788b109a629f9068/tests/test_cli.py#L89-L126

mrshu:
Thanks a lot for your comments @StrikerRUS!

I did add a test for the CLI part of the PR -- if you think it would be worth it, I am happy to add some for the "API" part as well.

m2cgen/cli.py
```diff
@@ -48,6 +49,10 @@
     choices=LANGUAGE_TO_EXPORTER.keys(),
     help="Target language",
     required=True)
+parser.add_argument(
+    "--function_name", "-fn", dest="function_name", type=str,
+    default="score",
+    help="Name of the function in the generated code (defaults to 'score')")
```
StrikerRUS: Seems that we haven't specified default values in docstrings for other arguments. Not sure that we should do it for this one.

m2cgen/exporters.py
```diff
@@ -16,6 +17,8 @@ def export_to_java(model, package_name=None, class_name="Model", indent=4):
         The name of the generated class.
     indent : int, optional
         The size of indents in the generated code.
+    function_name : string, optional
+        Name of the function in the generated code (defaults to 'score')
```
StrikerRUS: < Here and in all other places >
The same thing about writing default value in docstring as above. Also, please add dot at the end of sentence for the consistency with other descriptions.

m2cgen/exporters.py
```diff
@@ -158,17 +184,23 @@ def export_to_visual_basic(model, module_name="Model", indent=4):
...
+        module_name=module_name,
+        indent=indent,
+        function_name=function_name
+    )
     return _export(model, interpreter)
 
 
-def export_to_c_sharp(model, namespace="ML", class_name="Model", indent=4):
+def export_to_c_sharp(model, namespace="ML", class_name="Model", indent=4,
+                      function_name="Score"):
```
StrikerRUS: < Here and in all other places >
I believe it can be `score` for the consistency. Refer to https://github.com/BayesWitnesses/m2cgen/pull/166#issuecomment-586548983
> I'm not sure that it is worthy of further consideration. As we are still in alpha phase, we are free to change a lot without any deprecation warnings and worrying about the backward compatibility. Especially for just a naming conventions.

m2cgen/interpreters/c/interpreter.py
```diff
@@ -21,7 +21,9 @@ class CInterpreter(ToCodeInterpreter,
     power_function_name = "pow"
     tanh_function_name = "tanh"
 
-    def __init__(self, indent=4, *args, **kwargs):
+    def __init__(self, indent=4, function_name="score", *args, **kwargs):
+        self.function_name = function_name
```
StrikerRUS: It should be used later in 
```
with self._cg.function_definition(
        name=self.function_name,
```

m2cgen/interpreters/php/interpreter.py
```diff
@@ -20,7 +20,9 @@ class PhpInterpreter(ToCodeInterpreter, mixins.LinearAlgebraMixin):
     power_function_name = "pow"
     tanh_function_name = "tanh"
 
-    def __init__(self, indent=4, *args, **kwargs):
+    def __init__(self, indent=4, function_name="score", *args, **kwargs):
+        self.function_name = function_name
```
StrikerRUS: It should be used later.

izeigerman:
@StrikerRUS 

> To be honest, I don't remember I was saying this 😃
Oops, sorry. I confused 2 things 🤦‍♂

m2cgen/exporters.py
izeigerman: Hey @StrikerRUS shouldn't we at least comply with basic code style principles of a target language? Like in this case method names are capitalized in C#. I believe this basic logic is worth keeping.

m2cgen/cli.py
mrshu: I see, it is just something I would expect.

If you think it does not make sense to have it here, I am happy to get rid of it.

m2cgen/interpreters/c/interpreter.py
mrshu: My bad, I completely missed that.

Thanks for spotting this @StrikerRUS!

m2cgen/exporters.py
mrshu: I really do not have a strong opinion here, but since C# seems to be using [PascalCasing by definition](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/capitalization-conventions), I would vote for being consistent with that.

mrshu:
@StrikerRUS Thanks for your feedback!

I believe I've addressed most of the things you've pointed out except for the C# one where I'd like to wait for your opinion.

If you still see something worth updating here, feel free to let me know -- I'll gladly update it.

Thanks!

m2cgen/exporters.py
StrikerRUS: To be honest, I don't have a strong opinion on keeping code style for default values. I just proposed the way which is the easiest to maintain 🙂 .
And then we'll need to apply conditional default values for the CLI part: https://github.com/BayesWitnesses/m2cgen/pull/166#issuecomment-586503683.

m2cgen/cli.py
StrikerRUS: Thank you!

m2cgen/exporters.py
mrshu: @StrikerRUS All right, I went ahead and implemented the default values for the CLI part such that they directly use the default value of the `function_name` keyword argument.

It relies on inspecting the `export` function (the [inspect module](https://docs.python.org/3.5/library/inspect.html#inspect.signature) introduced in Python 3.5 so that should not be an issue) and might therefore feel a bit hacky, but it allows us to specify the default only once (namely in the keyword argument) and be done with it -- it should automatically down-propagate to the CLI options as well. 

I also added a test which should give us a greater confidence that this works.

If you do not like this approach, I am happy to revert that particular commit and just leave it as it was.

Thanks again!

StrikerRUS: @mrshu Thanks a lot for implementing conditional default values! I like your approach very much! Just please add a comment about that default value is conditional and will be set later based on the `export` function signature.
```
parser.add_argument(
    "--function_name", "-fn", dest="function_name", type=str,
    default=None,  # some comment here 
    help="Name of the function in the generated code.")
```

mrshu: Thanks @StrikerRUS , I just did that.

Please feel free to point out a simpler version of the comment -- I may have overcomplicated it quite a bit.

PR #167, opened 2020-02-15
Dart lang support

MattConflitti:
Issue of tanh function is still there. There is currently a unit test that expects the use of tanh to throw an error.

Edit: tanh issue is fixed thanks to guidance from @StrikerRUS


README.md
```diff
@@ -36,6 +36,7 @@ pip install m2cgen
 - Python
 - R
 - Visual Basic
+- Dart
```
StrikerRUS: Please save alphabetical order here. It will help users find (or not find 😄 ) their target language faster.

m2cgen/__init__.py
```diff
@@ -11,6 +11,7 @@
     export_to_powershell,
     export_to_r,
     export_to_php,
+    export_to_dart
```
StrikerRUS:
```suggestion
    export_to_dart,
```

As this list is supposed to be enhanced in the future, adding comma decreases future diffs and makes review process a little bit easier. Refer to https://stackoverflow.com/a/11597911.

m2cgen/__init__.py
```diff
@@ -24,6 +25,7 @@
...
```
StrikerRUS: ditto

m2cgen/cli.py
```diff
@@ -29,6 +29,7 @@
     "powershell": (m2cgen.export_to_powershell, ["indent"]),
     "r": (m2cgen.export_to_r, ["indent"]),
     "php": (m2cgen.export_to_php, ["indent"]),
+    "dart": (m2cgen.export_to_dart, ["indent", "class_name"])
```
StrikerRUS: ditto

m2cgen/interpreters/__init__.py
```diff
@@ -20,4 +21,5 @@
     PowershellInterpreter,
     RInterpreter,
     PhpInterpreter,
+    DartInterpreter
```
StrikerRUS: ditto for trailing comma

m2cgen/interpreters/dart/code_generator.py
```diff
@@ -0,0 +1,54 @@
...
+
+class DartCodeGenerator(CLikeCodeGenerator):
+
+    scalar_type = "double"
+    vector_type = "List<double>"
+
+    def __init__(self, *args, **kwargs):
+        super(DartCodeGenerator, self).__init__(*args, **kwargs)
+
+    def add_class_def(self, class_name, is_private=False):
```
StrikerRUS: I have never heard about the Dart language before 🙂 , but according to the following articles
- https://en.wikipedia.org/wiki/Dart_(programming_language)
- https://dart.dev/guides/language/language-tour

it seems to me that classes are optional stuff in this language. I think we should try to generate code as simple as possible and here it can be simplified to an ordinary function (unlike in Java and C# where classes are essential part of the language). Please correct me if I'm wrong.

m2cgen/interpreters/dart/tanh.dart
```diff
@@ -0,0 +1,7 @@
+static double tanh(double x) {
+    if (x > 22.0)
+        return 1.0;
+    if (x < -22.0)
+        return -1.0;
+    return ((exp(2*x) - 1)/(exp(2*x) + 1));
+}
```
StrikerRUS: Please add none blank line at the end of the file.

m2cgen/exporters.py
```diff
@@ -251,6 +251,29 @@ def export_to_php(model, indent=4):
     return _export(model, interpreter)
 
 
+def export_to_dart(model, class_name="Model", indent=4):
```
StrikerRUS: Please check the following comment: https://github.com/BayesWitnesses/m2cgen/pull/97#pullrequestreview-287153120. Feel free to ask for any help if you find something unclear in `e2e` tests.

README.md
MattConflitti: Shoot. Yes! My apologies.

m2cgen/interpreters/dart/code_generator.py
MattConflitti: I have only a little experience in Dart. I can definitely make this change. My only reservation with doing so is that class encapsulation makes sense as the linear algebra and tanh functions are added as helpers within the class and really do not need to be accessible outside (in which I could change the naming convention by prepending a '_' to keep them private). Also, it feels more natural accessing the RegressorModel class's functionality than just grabbing a function out of a file. If you could go either way, I would say the class method has a slight edge. At the end of the day, the developer using this will do whatever they please to the code to make it work in their project.

For what it is worth, Wikipedia defines the language as:
"Dart is an object-oriented, class defined, garbage-collected language..."
which sounds a lot like Java's definition:
"Java is a general-purpose programming language that is class-based, object-oriented..."

You let me know what the final say is, boss!

m2cgen/exporters.py
MattConflitti: Of course! I knew I missed something!

MattConflitti: One question I have is do these tests get run manually by a dev as a check or in the build pipeline of the package? My latest commit resulted in a successful build and it did not include the e2e tests. So, as long as the e2e DartExecutor passes on my system, that should suffice?

m2cgen/interpreters/dart/code_generator.py
StrikerRUS: Nowadays a lot of programming languages are multi-paradigm and we can say the same for many of them. For instance, we can use classes in JS, Python, PHP, PowerShell, etc. but we don't.

Anyway, @izeigerman will have the last word.

m2cgen/exporters.py
StrikerRUS: Once `DartExecutor` (with some corresponding changes) added to the package, new commits will be run at Travis with Dart e2e tests.

image

MattConflitti: I see that the code is running a command to test the script using the cli tool of that particular language. Do I need to provide information somewhere as to how to set up that cli tool in Travis? I'm afk right now so I don't have a chance to go through that directory to see if that exists.

m2cgen/interpreters/dart/code_generator.py
MattConflitti: I want to make sure I conform my code to what the contributors of this repo want, so I will do whatever you want.

I did come across this article for writing effective Dart code: https://dart.dev/guides/language/effective-dart/design#avoid-defining-a-class-that-contains-only-static-members

Looks like you have the upper hand 😉.

I will make the necessary changes as soon as I can.

MattConflitti: My next question would be, does using the Dart `library` construct mentioned in that article make sense? Could be a good way to scope the model or provide some semantic sugar upon importing the code.

I do not really know how the typical user uses the tool. I would guess devs will export the code and then rename things and paste the code where they want it as they see fit, so the extra code just causes headaches on our end to maintain it.

izeigerman: Thank you, folks. 
I'm a bit biased in this regard since I'm not a huge fan of the object-oriented paradigm and find classes to be redundant. Though in some languages they are the only way to achieve modularity, and hence a necessary evil.  

Long story short, since the code generated by `m2cgen` doesn't maintain any state that requires encapsulation plus Dart has first-class functions - I vote to drop classes. 

> My next question would be, does using the Dart library construct mentioned in that article make sense? Could be a good way to scope the model or provide some semantic sugar upon importing the code.

Quoting the source that you shared:
> If what you want is a namespace, a library is a better fit. Libraries support import prefixes and show/hide combinators. Those are powerful tools that let the consumer of your code handle name collisions in the way that works best for them.

So it seems like Dart leaves a lot to the consumer's consideration so why shouldn't we. Let's leave it up to a user to decide how to handle the generated code. Additionally having a function name parameterized should help with name collision. This feature is currently being introduced in the adjacent PR (https://github.com/BayesWitnesses/m2cgen/pull/166). Perhaps we should consider proactively introducing this parameter to the Dart interpreter.

izeigerman:
@MattConflitti thank you so much for this PR 🥇 💪

m2cgen/exporters.py
MattConflitti: @izeigerman or @StrikerRUS 
It appears the Dart language will throw an error if there is a copious amount of nesting in the code. This makes 3 tests fail, most likely because they are more complex models.

```
Error: The file has too many nested expressions or statements.
E           Try simplifying the code.
E               return ((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((((21.24023452121809) + ((exp((var0) * (((((((((((((pow((15.0234) - (input[0]), 2)) + (pow((0.0) - (input[1]), 2))) + (pow((18.1) - (input[2]), 2))) + (pow((0.0) - (input[3]), 2))) + (pow((0.614) - (input[4]), 2))) + (pow((5.304) - (input[5]), 2))) + (pow((97.3) - (input[6]), 2))) + (pow((2.1007)...
```
Screen Shot 2020-02-18 at 8 46 44 PM

Any thoughts? Obviously, there is no reason to change the interpreter, but is there a way to silence these tests and if so, is that permitted? This is a constraint on the language not m2cgen.

StrikerRUS: @MattConflitti Thank you very much for setting up e2e tests for Dart language! You did it without any guidance from us, impressive! You rock!

Speaking about 3 failing tests, it seems that you've encountered the same problem as me for R language some time ago. Please refer to https://github.com/BayesWitnesses/m2cgen/pull/139#issuecomment-568942961. As a workaround, Dart interpreter should inherit `mixins.BinExpressionDepthTrackingMixin` class and you should play around with the default value for `bin_depth_threshold` which will allow to generate all types of models with reasonable complexity. Also, refer to https://github.com/BayesWitnesses/m2cgen/pull/139#discussion_r361502593.

MattConflitti: Great! I will make that change and see if I can get it to pass! I appreciate the guidance!

.travis/setup.sh
```diff
@@ -29,3 +29,13 @@ if [[ $LANG == *"php"* ]]; then
   sudo apt-get update
   sudo apt-get install --no-install-recommends -y php
 fi
+
+# Install Dart. (https://dart.dev/get-dart)
+if [[ $LANG == *"dart"* ]]; then
+  sudo apt-get update
+  sudo apt-get install apt-transport-https
```
StrikerRUS: I believe Travis has already installed that package and we can save some CI time by removing these lines.
```
The following packages will be upgraded:
  apt-transport-https
1 upgraded, 0 newly installed, 0 to remove and 253 not upgraded.
Need to get 26.5 kB of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 http://archive.ubuntu.com/ubuntu xenial-updates/main amd64 apt-transport-https amd64 1.2.32 [26.5 kB]
Fetched 26.5 kB in 0s (89.4 kB/s)
(Reading database ... 126205 files and directories currently installed.)
Preparing to unpack .../apt-transport-https_1.2.32_amd64.deb ...
Unpacking apt-transport-https (1.2.32) over (1.2.31) ...
Setting up apt-transport-https (1.2.32) ...
```

m2cgen/exporters.py
```diff
@@ -251,6 +251,29 @@ def export_to_php(model, indent=4):
...
+    """
+    Generates a Dart code representation of the given model.
+
+    Parameters
+    ----------
+    model : object
+        The model object that should be transpiled into code.
+    class_name : string, optional
+        The name of the generated class.
```
StrikerRUS: As we've dropped classes, I think it can be removed now.

m2cgen/exporters.py
MattConflitti: It appears to have worked. Should I use the subroutine mixin as well? If I understand it correctly, it is essentially building proxy functions to call so that there is less nesting? Is that necessary if the bin threshold works up to 465 for Dart? Thanks!

izeigerman: > Should I use the subroutine mixin as well?

No as long as there are no restrictions on a size of the individual function/method in DART.

> Is that necessary if the bin threshold works up to 465 for Dart? 

Not really. `BinExpressionDepthTrackingMixin ` and the Subroutine mixin address somewhat different problems. The former workarounds the constraints on the depth of binary expressions while the later helps overcome function/method size constraints.

MattConflitti:
Great! Then the tests pass and this PR should be pretty much set as far
as I can tell! Thanks for your help.

tests/interpreters/test_dart.py
```diff
@@ -0,0 +1,293 @@
+from m2cgen import ast
```
StrikerRUS: Please add some more tests related to the `BinExpressionDepthTrackingMixin`. You can use the following code as example:
https://github.com/BayesWitnesses/m2cgen/blob/f14114614113d018531b7232750cc6ed977272e8/tests/interpreters/test_r.py#L220-L361
https://github.com/BayesWitnesses/m2cgen/blob/f14114614113d018531b7232750cc6ed977272e8/tests/interpreters/test_python.py#L187-L312
https://github.com/BayesWitnesses/m2cgen/blob/f14114614113d018531b7232750cc6ed977272e8/tests/interpreters/test_java.py#L355-L521

MattConflitti:
@StrikerRUS @izeigerman  Thank you for your help and maintaining this repo!

StrikerRUS:
@MattConflitti Can you please generate [code examples](https://github.com/BayesWitnesses/m2cgen/tree/master/generated_code_examples) for Dart in a follow-up PR?
Add Dart here
https://github.com/BayesWitnesses/m2cgen/blob/3c431f2c91c55cffe075c600efdcd9730465918c/tools/generate_code_examples.py#L34-L45
and run the following comand
https://github.com/BayesWitnesses/m2cgen/blob/3c431f2c91c55cffe075c600efdcd9730465918c/Makefile#L15

MattConflitti:
Absolutely!

PR #172, opened 2020-02-27
added support for lightning

StrikerRUS:
lightning is a library with various linear ML models implemented in Cython. It is listed in scikit-learn related projects: https://scikit-learn.org/stable/related_projects.html.

https://github.com/scikit-learn-contrib/lightning
http://contrib.scikit-learn.org/lightning/references.html

Not included models in this PR:
- [`KernelSVC`](http://contrib.scikit-learn.org/lightning/generated/lightning.classification.KernelSVC.html#lightning.classification.KernelSVC): I'll add it in a separate PR later
- [`SGDRegressor`](http://contrib.scikit-learn.org/lightning/generated/lightning.regression.SGDRegressor.html#lightning.regression.SGDRegressor): produces assertion error `expected=59.41283828752698, actual=60.700744525892105`
  **UPD:** I guess they forgot to add `intercept`, thus assertion error for default params where `fit_intercept=True`:
  https://github.com/scikit-learn-contrib/lightning/blob/b96f9c674968496e854078163c8814049a7b9f43/lightning/impl/sgd.py#L382-L406
cf. https://github.com/scikit-learn-contrib/lightning/blob/b96f9c674968496e854078163c8814049a7b9f43/lightning/impl/base.py#L107-L116
- [`SVRGClassifier`](http://contrib.scikit-learn.org/lightning/generated/lightning.classification.SVRGClassifier.html#lightning.classification.SVRGClassifier) / [`SVRGRegressor`](http://contrib.scikit-learn.org/lightning/generated/lightning.regression.SVRGRegressor.html#lightning.regression.SVRGRegressor): training always results in all `NaN`s in prediction results. Maybe there are some dependency conflicts or compilation errors, but the same behavior I observe on my local Windows machine with precompiled package from conda.


.travis.yml
```diff
@@ -16,6 +16,7 @@ before_install:
   - bash .travis/setup.sh
 
 install:
+  - pip install Cython
```
StrikerRUS: Why not include in `requirements-test.txt`: https://github.com/h5py/h5py/issues/535#issuecomment-74833863

m2cgen/assemblers/__init__.py
```diff
@@ -25,75 +25,128 @@
...
+    "sklearn.ensemble.forest.ExtraTreesRegressor":
+        RandomForestModelAssembler,
+    "sklearn.ensemble.forest.RandomForestClassifier":
+        RandomForestModelAssembler,
+    "sklearn.ensemble.forest.RandomForestRegressor":
+        RandomForestModelAssembler,
 }
 
 
+def _get_full_model_name(model):
```
StrikerRUS: Idea was taken from https://github.com/BayesWitnesses/m2cgen/pull/104#issuecomment-544703303. There are different models with same class names in different packages and we need to distinguish them. For example, we support `SGDRegressor` from scikit-learn and we do not support `SGDRegressor` from lightning.

requirements-test.txt
```diff
@@ -1,4 +1,4 @@
-numpy==1.15.1
+numpy==1.16.1
```
StrikerRUS: Fix the following error for Python 3.6:
```
ImportError while loading conftest '/home/travis/build/BayesWitnesses/m2cgen/tests/conftest.py'.
tests/conftest.py:3: in <module>
    from tests import utils
tests/utils.py:13: in <module>
    from lightning.classification import (
../../../virtualenv/python3.6.7/lib/python3.6/site-packages/lightning/classification.py:1: in <module>
    from .impl.adagrad import AdaGradClassifier
../../../virtualenv/python3.6.7/lib/python3.6/site-packages/lightning/impl/adagrad.py:10: in <module>
    from .base import BaseClassifier, BaseRegressor
../../../virtualenv/python3.6.7/lib/python3.6/site-packages/lightning/impl/base.py:14: in <module>
    from .randomkit import RandomState
../../../virtualenv/python3.6.7/lib/python3.6/site-packages/lightning/impl/randomkit/__init__.py:1: in <module>
    from .random_fast import RandomState
__init__.pxd:918: in init lightning.impl.randomkit.random_fast
    ???
E   ValueError: numpy.ufunc size changed, may indicate binary incompatibility. Expected 216 from C header, got 192 from PyObject
```

Refer to https://github.com/scikit-learn-contrib/hdbscan/issues/272#issuecomment-484383051.

requirements-test.txt
```diff
@@ -10,3 +10,4 @@ coveralls==1.9.2
 pytest-cov==2.8.1
 py-mini-racer==0.1.18
 statsmodels==0.10.2
+git+git://github.com/scikit-learn-contrib/lightning.git@b96f9c674968496e854078163c8814049a7b9f43
```
StrikerRUS: 0.5.0 from PyPI fails with Python 3.7.
Refer to https://github.com/scikit-learn-contrib/lightning/pull/133, https://github.com/scikit-learn-contrib/lightning/issues/139, https://github.com/scikit-learn-contrib/lightning/issues/132.

m2cgen/assemblers/__init__.py
```diff
@@ -25,75 +25,128 @@
...
+    type_name = type(model)
+    return "{}.{}".format(type_name.__module__,
```
StrikerRUS: Not sure, maybe we should take only the first part of `__module__` to be more robust for internal changes in libraries?.. Like `sklearn.DecisionTreeClassifier` instead of `sklearn.tree.tree.DecisionTreeClassifier` and `lightning.CDClassifier` instead of `lightning.impl.primal_cd.CDClassifier`.

izeigerman: I think I like this idea very much. Taking the first part will indeed help us to be independent from library internals. The only risk I'm seeing here is a name collision within the same library, which I hope is not something that happens often in the Python world. I'd only suggest to use a separator other than `.` like eg. `_` or `-` so that identifiers are not confused with actual package names. Eg. `sklearn.LogisticRegression` vs `sklearn_LogisticRegression`. Great idea otherwise 👍

izeigerman:
Hey @StrikerRUS, fantastic update which should also unblock the https://github.com/BayesWitnesses/m2cgen/pull/104. I think we should go for that idea that you suggested in comments. LGTM otherwise 👍

m2cgen/assemblers/linear.py
```diff
@@ -33,7 +33,8 @@ def _get_coef(self):
 class SklearnLinearModelAssembler(BaseLinearModelAssembler):
 
     def _get_intercept(self):
-        return self.model.intercept_
+        return getattr(self.model, "intercept_",
```
izeigerman: Is it worth to have a `LightingLinearModelAssembler` version of this assembler which inherits `SklearnLinearModelAssembler` and overrides the `_get_intercept` method with:
```
def _get_intercept(self):
    return np.zeros(self._get_coef().shape[0])
```
This way our expectations to those libraries will be clearly defined.

m2cgen/assemblers/__init__.py
StrikerRUS: Thanks a lot for your opinion! I absolutely agree with you about possible collisions. Made update in the latest commit.

m2cgen/assemblers/linear.py
StrikerRUS: It will override this method in the following way:
```
    def _get_intercept(self):
        return getattr(self.model, "intercept_",
                       np.zeros(self._get_coef().shape[0]))
```
In `lightning`, model can have `intercept_` attribute and can have no this attribute. In `scikit-learn`, model has `intercept_` attribute in all cases. Please tell me whether you still want to introduce `LightingLinearModelAssembler` or not after the info above.

izeigerman: Ah, I see. Thanks for providing this context. Let's keep it as is then.

StrikerRUS:
Thank you too! 🙂 

For the record: support for the [`SGDRegressor`](http://contrib.scikit-learn.org/lightning/generated/lightning.regression.SGDRegressor.html#lightning.regression.SGDRegressor) model can be added in the future after fixing the following [upstream issue](https://github.com/scikit-learn-contrib/lightning/issues/141).

PR #176, opened 2020-03-08
added lightning.KernelSVC (binary only)

StrikerRUS:
Refer to https://github.com/BayesWitnesses/m2cgen/pull/172#issue-380577165. [`KernelSVC`](http://contrib.scikit-learn.org/lightning/generated/lightning.classification.KernelSVC.html) is the last model from `lightning` library that can be added to `m2cgen`.

Adding only binary classification to not make PR too huge.


m2cgen/assemblers/svm.py
```diff
@@ -135,3 +170,39 @@ def _linear_kernel_with_gama_and_coef(self, support_vector):
...
+        return self.model.coef_[0]
+
+    def _cosine_kernel(self, support_vector):
+        support_vector_norm = np.linalg.norm(support_vector)
+        feature_norm = ast.SqrtExpr(
+            utils.apply_op_to_expressions(
+                ast.BinNumOpType.ADD,
+                *[utils.mul(ast.FeatureRef(i), ast.FeatureRef(i))
+                  for i in range(len(support_vector))]),
+            to_reuse=True)
```
StrikerRUS: Have no idea how to avoid zero norm 🙁

m2cgen/assemblers/svm.py
```diff
@@ -57,6 +46,53 @@ def _assemble_single_output(self):
...
+
+    def _get_single_intercept(self):
+        raise NotImplementedError
+
+
+class SklearnSVMModelAssembler(BaseSVMModelAssembler):
+
+    def _get_supported_kernels(self):
+        return {
+            "rbf": self._rbf_kernel,
```
izeigerman: Since all 4 of these kernels are shared between implementations and reside in the base class does it make sense to move the body of this `_get_supported_kernels` into the base class and make it a default behavior?

m2cgen/ast.py
```diff
@@ -237,7 +249,7 @@ def __str__(self):
     (PowExpr, lambda e: [e.base_expr, e.exp_expr]),
     (VectorVal, lambda e: e.exprs),
     (IfExpr, lambda e: [e.test, e.body, e.orelse]),
-    ((ExpExpr, TanhExpr, TransparentExpr), lambda e: [e.expr]),
+    ((ExpExpr, SqrtExpr, TanhExpr, TransparentExpr), lambda e: [e.expr]),
```
izeigerman: Do we need to include `SqrtExpr` into one of the tests that we have for `count_exprs`?

README.md
```diff
@@ -43,7 +43,7 @@ pip install m2cgen
...
-| **SVM** | <ul><li>scikit-learn<ul><li>LinearSVC</li><li>NuSVC</li><li>SVC</li></ul></li><li>lightning<ul><li>LinearSVC</li></ul></li></ul> | <ul><li>scikit-learn<ul><li>LinearSVR</li><li>NuSVR</li><li>SVR</li></ul></li><li>lightning<ul><li>LinearSVR</li></ul></li></ul> |
+| **SVM** | <ul><li>scikit-learn<ul><li>LinearSVC</li><li>NuSVC</li><li>SVC</li></ul></li><li>lightning<ul><li>KernelSVC</li><li>LinearSVC</li></ul></li></ul> | <ul><li>scikit-learn<ul><li>LinearSVR</li><li>NuSVR</li><li>SVR</li></ul></li><li>lightning<ul><li>LinearSVR</li></ul></li></ul> |
```
izeigerman: Do we need to mentioned that the multi-class is not supported?

m2cgen/assemblers/svm.py
izeigerman: What is the lighting's behavior for zero future norm? Can we get away with the same logic that we use for Support Vector norm?

StrikerRUS: It relies on scikit-learn code:
https://github.com/scikit-learn/scikit-learn/blob/95d4f0841d57e8b5f6b2a570312e9d832e69debc/sklearn/preprocessing/_data.py#L1730
https://github.com/scikit-learn/scikit-learn/blob/95d4f0841d57e8b5f6b2a570312e9d832e69debc/sklearn/preprocessing/_data.py#L63-L78

m2cgen/assemblers/svm.py
StrikerRUS: Yeah, as we have only 2 implementations of SVM for now, I think it makes sense.

m2cgen/ast.py
StrikerRUS: Sure! Forgot about it, sorry!

m2cgen/assemblers/svm.py
izeigerman: If I understand the logic correctly it does the same thing (falls back to 1.0), shall we do the same here?

izeigerman:
Let's add a fallback logic for the feature norm and this is good to go. Thanks for addressing comments 👍

m2cgen/assemblers/svm.py
StrikerRUS: To be honest, I'm not sure that complicating generated code in a such way worth it...

izeigerman: I'd say we should always prefer correctness even if it means a slightly more complex code. Btw, does it really become that complex? Especially considering that we reuse the expression (to_reuse=True) here. AST doesn't look bad at all. Thanks for implementing this! 👍

tests/assemblers/test_svm.py
```diff
@@ -117,11 +117,23 @@ def kernel_ast(sup_vec_value):
...
+                        ast.SqrtExpr(
+                            ast.BinNumExpr(
+                                ast.FeatureRef(0),
+                                ast.FeatureRef(0),
+                                ast.BinNumOpType.MUL),
+                            to_reuse=True),
+                        ast.NumVal(0.0),
+                        ast.CompOpType.EQ),
+                    ast.NumVal(1.0),
+                    ast.SqrtExpr(
```
izeigerman: [Nitpick] This part can be extracted into a separate variable and cached this way making expected AST a little simpler:
```
feature_norm = ast.SqrtExpr(
                            ast.BinNumExpr(
                                ast.FeatureRef(0),
                                ast.FeatureRef(0),
                                ast.BinNumOpType.MUL),
                            to_reuse=True)

...
ast.IfExpr(
                    ast.CompExpr(
                        feature_norm,
                        ast.NumVal(0.0),
                        ast.CompOpType.EQ),
                    ast.NumVal(1.0),
                    feature_norm)
```

m2cgen/assemblers/svm.py
StrikerRUS: > we should always prefer correctness even if it means a slightly more complex code. 

Yeah, no doubt!

In languages which implement `to_reuse=True` it looks OK, but if they don't... Though, there are many other places where without caching code looks terrible.

I thought about more elegant solution, but didn't come to any.

tests/assemblers/test_svm.py
StrikerRUS: Sure, will do this in a separate PR.

m2cgen/assemblers/svm.py
izeigerman: > I thought about more elegant solution, but didn't come to any.
Well, TBH, the existing solution looks more than good to me at least at this point 👍

PR #178, opened 2020-03-10
Generalize the Subroutine approach implemented in the Java interpreter and use it in the R interpreter

izeigerman:
Finally got a chance to implement this. I have updated the Subroutine Mixin to use the logic I earlier implemented in the Java interpreter related to splitting the code into methods based on the AST size. Did minor refactoring and renaming there. The R interpreter now uses this approach as well.

Will be dropping the `SubroutineExpr` expression from AST in a follow-up PR.


Dockerfile
```diff
@@ -29,7 +32,9 @@ RUN apt-get update && \
 WORKDIR /m2cgen
 
 COPY requirements-test.txt ./
-RUN pip3 install --no-cache-dir Cython && \
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 1 && \
+    pip3 install --upgrade pip && \
+    pip3 install --no-cache-dir Cython numpy && \
```
izeigerman: `numpy` is being used as part of a `setup.py` in either statsmodels or lighting (don't remember which one, can double check if needed), that's why we adding it here.

Dockerfile
```diff
@@ -29,7 +32,9 @@ RUN apt-get update && \
 WORKDIR /m2cgen
 
 COPY requirements-test.txt ./
-RUN pip3 install --no-cache-dir Cython && \
+RUN update-alternatives --install /usr/bin/python3 python3 /usr/bin/python3.7 1 && \
```
izeigerman: Turns out the image had been using Python 3.5 all this time.

Dockerfile
```diff
@@ -11,12 +11,15 @@ RUN apt-get update && \
     wget -qO- https://storage.googleapis.com/download.dartlang.org/linux/debian/dart_stable.list > /etc/apt/sources.list.d/dart_stable.list && \
     apt-get update && \
     apt-get install --no-install-recommends -y \
+        git \
```
izeigerman: These new packages a required to successfully install statsmodels and lighting Python packages. The image was broken all this time.

m2cgen/interpreters/r/interpreter.py
```diff
@@ -18,6 +17,8 @@ class RInterpreter(ToCodeInterpreter,
     # This value is just a heuristic and is subject to change in the future
     # based on the users' feedback.
     bin_depth_threshold = 25
+    ast_size_check_frequency = 2
```
izeigerman: Depending on a Travis build additional tuning may be required. The execution time looks reasonable when running locally.

izeigerman:
@StrikerRUS I think this one is ready to be reviewed and discussed.

StrikerRUS:
Huh, wanted to review this PR, but GitHub said "Not this time!"

Error in browser console: `0: unable to load compat.js`.

I even cannot attach the screenshot with this error and post a comment in any thread.

Dockerfile
StrikerRUS: Ooops, I'm very sorry that I haven't done the needed updates in corresponding PRs!

Makefile
```diff
@@ -15,7 +15,7 @@ docker-generate-examples:
 	$(DOCKER_RUN_ARGS) bash -c "python3 setup.py develop && python3 tools/generate_code_examples.py generated_code_examples"
 
 docker-flake8:
-	$(DOCKER_RUN_ARGS) bash -c "flake8 ."
+	$(DOCKER_RUN_ARGS) bash -c "flake8 --exclude tmp,generated_code_examples ."
```
StrikerRUS: Maybe sync with 
https://github.com/BayesWitnesses/m2cgen/blob/e0bae16a64c2ac28abd8fe64f1c926c2f8a539db/setup.cfg#L6-L7
?

izeigerman: Hm, weird. Not sure how I overlooked that and why the config wasn't picked up in my local env. Thanks for pointing this out

PR #185, opened 2020-03-22
update code for XGBoost to support newer versions

### LINKED ISSUE #183: Code generated from XGBoost model includes "None"
When transpiling `XGBRegressor` and `XGBClassifier` models such as the following basic example:

```python
from xgboost import XGBRegressor
from sklearn import datasets
import m2cgen as m2c

iris_data = datasets.load_iris(return_X_y=True)

mod = XGBRegressor(booster="gblinear", max_depth=2)
X, y = iris_data
mod.fit(X[:120], y[:120])

code = m2c.export_to_c(mod)

print(code)
```

the resulting c-code includes a Pythonesque `None` :

```c
double score(double * input) {
    return (None) + (((((-0.391196) + ((input[0]) * (-0.0196191))) + ((input[1]) * (-0.11313))) + ((input[2]) * (0.137024))) + ((input[3]) * (0.645197)));
}
```

Probably I am missing some basic step?

StrikerRUS:
Fixed #183.

Refer to 
https://github.com/dmlc/xgboost/blob/cfae247231e4977a78b255810565369c934fe1b5/tests/python/test_with_sklearn.py#L535-L537

### DIFF
--- a/tests/assemblers/test_xgboost.py
+++ b/tests/assemblers/test_xgboost.py
@@ -263,7 +263,9 @@ def test_regression_saved_without_feature_names():
 
 
 def test_linear_model():
+    # Default updater ("shotgun") is nondeterministic
     estimator = xgboost.XGBRegressor(n_estimators=2, random_state=1,
+                                     updater="coord_descent",
                                      feature_selector="shuffle",
                                      booster="gblinear")
     utils.get_regression_model_trainer()(estimator)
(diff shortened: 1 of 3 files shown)


tests/assemblers/test_xgboost.py
```diff
@@ -247,23 +247,25 @@ def test_regression_saved_without_feature_names():
...
-            ast.NumVal(1.71813202)),
+            ast.NumVal(8.34557438),
+            ast.NumVal(3.9141891)),
         ast.BinNumOpType.ADD)
 
     assert utils.cmp_exprs(actual, expected)
 
 
 def test_linear_model():
+    # Default updater ("shotgun") is nondeterministic
```
StrikerRUS: Refer to
> updater [default= shotgun]
       Choice of algorithm to fit linear model
           shotgun: Parallel coordinate descent algorithm based on shotgun algorithm. Uses ‘hogwild’ parallelism and therefore produces a nondeterministic solution on each run.
https://xgboost.readthedocs.io/en/latest/parameter.html?highlight=shotgun#parameters-for-linear-booster-booster-gblinear

and
> Using gblinear booster with shotgun updater is nondeterministic as
           it uses Hogwild algorithm.
https://github.com/dmlc/xgboost/blob/cd7d6f7d596abfe8eb81651386d8c05507c28fc8/python-package/xgboost/sklearn.py#L105-L106

It causes different results in API test which leads to impossibility to write `expected` expression.

izeigerman:
Thanks for fixing this so swiftly 👍Is the new method to extract the `base_core` compatible with older versions of XGBoost (eg. 0.90)?

StrikerRUS:
Yes, I tested this method with `0.90` version. Also, in theory it should be compatible with all versions as `get_params()` is the method from fundamental scikit-learn's `BaseEstimator` class. But I'm not sure that XGBoost follows scikit-learn's dev requirements fully.

https://scikit-learn.org/stable/developers/develop.html#get-params-and-set-params

PR #188, opened 2020-03-30
Fix #168. Enforce float32 type for split condition values for GBT models created using XGBoost

### LINKED ISSUE #168: Code generated for XGBoost models returns invalid scores when tree_method is set to "hist"
I have trained xgboost models in Python and am using the CLI interface to convert the serialized models to pure python.  However, when I use the pure python, the results differ from the predictions using the model directly.

Python 3.7
xgboost 0.90

My model has a large number of parameters (somewhat over 500).
Here are predicted class probabilities from the original model:
image

Here are the same predicted probabilities using the generated python code via m2cgen:
image

We can see that the results are similar but not the same.  The result is a significant number of cases that are moved into different classes between the two sets of predictions.

I have also tested this with binary classification models and have the same issues.

izeigerman:
As it turns out the issue reported in https://github.com/BayesWitnesses/m2cgen/issues/168 is not unique to the "hist" tree construction algorithm. It seems that with "hist" method the likelihood of reprdocue is much higher due to relying on feature histograms. I was able to reproduce the same discrepancy with non-hist methods on a larger sample of test data.

The issue occurs due to a double precision error and reproduces every time when the feature value matches the split condition in one of the tree's nodes. 

Example: feature value = `0.671`, split condition = `0.671000004`. When we hit this condition in the generated code the outcome of `0.671 <  0.671000004` is "true" (or "yes" branch). While in XGBoost the same condition leads to the "no" branch.

 After some investigation I noticed that the XGBoost's `DMatrix` forces all values to be `float32` (https://github.com/dmlc/xgboost/blob/master/python-package/xgboost/core.py#L565). At the same time in our assemblers we rely on default 64-bit floats. Forcing the split condition to be `float32` seem to address the issue. At least I couldn't reproduce it so far.

### DIFF
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -134,7 +134,7 @@ class XGBoostTreeModelAssembler(BaseTreeBoostingAssembler):
         if "leaf" in tree:
             return ast.NumVal(tree["leaf"])
 
-        threshold = ast.NumVal(np.float32(tree["split_condition"]))
+        threshold = ast.NumVal(tree["split_condition"], dtype=np.float32)
         split = tree["split"]
         feature_idx = self._feature_name_to_idx.get(split, split)
         feature_ref = ast.FeatureRef(feature_idx)
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -134,7 +134,7 @@ class XGBoostTreeModelAssembler(BaseTreeBoostingAssembler):
         if "leaf" in tree:
             return ast.NumVal(tree["leaf"])
 
-        threshold = ast.NumVal(tree["split_condition"])
+        threshold = ast.NumVal(np.float32(tree["split_condition"]))
         split = tree["split"]
         feature_idx = self._feature_name_to_idx.get(split, split)
         feature_ref = ast.FeatureRef(feature_idx)
--- a/tests/e2e/test_e2e.py
+++ b/tests/e2e/test_e2e.py
@@ -35,50 +35,51 @@ CLASSIFICATION = pytest.mark.clf
 
 
 # Set of helper functions to make parametrization less verbose.
-def regression(model):
+def regression(model, test_fraction=0.02):
     return (
         model,
-        utils.get_regression_model_trainer(),
+        utils.get_regression_model_trainer(test_fraction),
         REGRESSION,
     )
 
 
-def classification(model):
+def classification(model, test_fraction=0.02):
     return (
         model,
-        utils.get_classification_model_trainer(),
+        utils.get_classification_model_trainer(test_fraction),
         CLASSIFICATION,
     )
 
 
-def classification_binary(model):
+def classification_binary(model, test_fraction=0.02):
     return (
         model,
-        utils.get_binary_classification_model_trainer(),
+        utils.get_binary_classification_model_trainer(test_fraction),
         CLASSIFICATION,
     )
 
 
-def regression_random(model):
+def regression_random(model, test_fraction=0.02):
     return (
         model,
-        utils.get_regression_random_data_model_trainer(0.01),
+        utils.get_regression_random_data_model_trainer(test_fraction),
         REGRESSION,
     )
 
 
-def classification_random(model):
+def classification_random(model, test_fraction=0.02):
     return (
         model,
-        utils.get_classification_random_data_model_trainer(0.01),
+        utils.get_classification_random_data_model_trainer(test_fraction),
         CLASSIFICATION,
     )
 
 
-def classification_binary_random(model):
+def classification_binary_random(model, test_fraction=0.02):
     return (
         model,
-        utils.get_classification_binary_random_data_model_trainer(0.01),
+        utils.get_classification_binary_random_data_model_trainer(
+            test_fraction),
         CLASSIFICATION,
     )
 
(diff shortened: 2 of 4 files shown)


izeigerman:
@StrikerRUS In case if you're curious to play with this issue and for the historical purposes I'm attaching a serialized (pickle) trained XGBoost regression model (trained using the "hist" method). 
The feature vector on which the issue reproduces:
```
[9.82349, 0.0, 18.1, 0.0, 0.671, 6.794, 98.8, 1.358, 24.0, 666.0, 20.2, 396.9, 10.17]
```
Note the `0.671` value (feature `f4`). Because of this value the discrepancy occurs in the estimator with index 7, node ID 4:
```
"nodeid": 4, "depth": 2, "split": "f4", "split_condition": 0.671000004, "yes": 9, "no": 10, "missing": 10
```
This is where the generated code follows the "yes" path while XGBoost does the opposite.
[round_error_xgboost.bin.gz](https://github.com/BayesWitnesses/m2cgen/files/4400219/round_error_xgboost.bin.gz)

StrikerRUS:
Ah, brilliant investigation!
I completely forgot about that we can encounter the same issue as https://github.com/dmlc/treelite/issues/95, https://github.com/dmlc/treelite/pull/55.

BTW, doesn't this code contradict the following from one of treelite's issue?
https://github.com/BayesWitnesses/m2cgen/blob/52c601b9d296c40ac0f946d89e08cbdf3bd931f4/m2cgen/assemblers/tree.py#L54-L57
> One of the model builder tests failed because scikit-learn uses double-precision floating-point to store split points, whereas Treelite was using single-precision.

Also refer to 
https://github.com/scikit-learn/scikit-learn/blob/95d4f0841d57e8b5f6b2a570312e9d832e69debc/sklearn/tree/_tree.pyx#L544-L545
and
https://github.com/scikit-learn/scikit-learn/blob/95d4f0841d57e8b5f6b2a570312e9d832e69debc/sklearn/tree/_tree.pyx#L727

m2cgen/assemblers/boosting.py
```diff
@@ -134,7 +134,7 @@ def _assemble_tree(self, tree):
         if "leaf" in tree:
             return ast.NumVal(tree["leaf"])
 
-        threshold = ast.NumVal(tree["split_condition"])
+        threshold = ast.NumVal(np.float32(tree["split_condition"]))
```
StrikerRUS: Maybe a more general solution will be to add an optional `dtype` constructor argument? I mean,
```
class NumVal(NumExpr):
    def __init__(self, value, dtype=np.float64):
```

tests/e2e/test_e2e.py
```diff
@@ -35,50 +35,51 @@
...
     return (
         model,
-        utils.get_binary_classification_model_trainer(),
+        utils.get_binary_classification_model_trainer(test_fraction),
         CLASSIFICATION,
     )
 
 
-def regression_random(model):
+def regression_random(model, test_fraction=0.02):
```
StrikerRUS: Does `test_fraction` increase for random datasets allow to reproduce the original issue?

StrikerRUS:
For the reference: a more explicit proof that thresholds are supposed to be float:

https://github.com/dmlc/xgboost/blob/ab7a46a1a4b880607a965f78a9703a56de9915a3/src/tree/tree_model.cc#L248-L253

https://github.com/dmlc/xgboost/blob/ab7a46a1a4b880607a965f78a9703a56de9915a3/include/xgboost/tree_model.h#L150-L153

https://github.com/dmlc/xgboost/blob/ab7a46a1a4b880607a965f78a9703a56de9915a3/include/xgboost/tree_model.h#L103


Also, it seems that threshold can be integer:
https://github.com/dmlc/xgboost/blob/ab7a46a1a4b880607a965f78a9703a56de9915a3/src/tree/tree_model.cc#L229-L239

m2cgen/assemblers/boosting.py
izeigerman: That's a good idea 👍

tests/e2e/test_e2e.py
izeigerman: Yes, unfortunately the default fraction produced way too few samples to be able to reproduce the issue reliably.

izeigerman:
@StrikerRUS Thanks for all the additional context! 

> scikit-learn uses double-precision floating-point to store split points

That's rather weird. I clearly remember that scikit-learn used float32 in its tree implementation as well. Perhaps a more recent fix?

tests/e2e/test_e2e.py
StrikerRUS: Got it! We definitely need to refactor testing routines to be more tunable, e.g. allow to adjust `test_fraction` according to programming language. Refer to https://github.com/BayesWitnesses/m2cgen/pull/114#issuecomment-566822606.

StrikerRUS:
> I clearly remember that scikit-learn used float32 in its tree implementation as well. 

Have no idea...
Version `0.18` (2016). Thresholds are `double`.

https://github.com/scikit-learn/scikit-learn/blob/38030a00a7f72a3528bd17f2345f34d1344d6d45/sklearn/tree/_tree.pyx#L186
(was later removed as unused in https://github.com/scikit-learn/scikit-learn/pull/13230)

https://github.com/scikit-learn/scikit-learn/blob/38030a00a7f72a3528bd17f2345f34d1344d6d45/sklearn/tree/_tree.pyx#L527-L528

izeigerman:
@StrikerRUS Got a chance to address your comment only now. Sorry about the delay.

StrikerRUS:
No problem at all! 🙂 

Seems that linter is unhappy with imported `numpy` now. Also, I think we can have a new API test to ensure that passing different `dtype`s really takes effect.

For scikit-learn issue (https://github.com/BayesWitnesses/m2cgen/pull/188#issuecomment-607403434) I believe it is better to have a separate PR.

izeigerman:
Fixed the linter error and added a test. Thanks 👍

PR #190, opened 2020-04-05
Use NumVal with specified 'dtype' instead of converting the threshold value in place when assembling a sklearn tree

izeigerman:


izeigerman:
@StrikerRUS Based on our discussion earlier I suggest we merge this PR unless you have objections.

PR #191, opened 2020-04-11
support Python 3.8

StrikerRUS:
Test requirements versions bump was needed because old packages are not compiled for Python 3.8.

**UPD:** BTW, what do you think about setting a dependency bot which will create PRs when new versions are released? I believe it is quite important to be sure that we are compatible with the latest versions of libraries we are supporting.

It looks like the following

image


tests/utils.py
```diff
@@ -218,7 +218,7 @@ def predict_from_commandline(exec_args):
     result = subprocess.Popen(exec_args, stdout=subprocess.PIPE,
                               stderr=subprocess.PIPE)
     stdout, stderr = result.communicate()
-    if result.returncode is not 0:
+    if result.returncode != 0:
```
StrikerRUS, line 221:
```
./tests/utils.py:221:8: F632 use ==/!= to compare str, bytes, and int literals
```

izeigerman:
Awesome, thanks for keeping the project's env up to date 👍

StrikerRUS:
No problem! It was necessary to be compatible with Python 3.8.
When have time, please take a brief glance at dependency bots and drop some words about what do you think about them here.

StrikerRUS:
Gently ping @izeigerman for thoughts about dependency bot.

PR #195, opened 2020-04-16
added statsmodels GLM

StrikerRUS:
https://www.statsmodels.org/stable/generated/statsmodels.genmod.generalized_linear_model.GLM.html


m2cgen/assemblers/linear.py
```diff
@@ -70,6 +75,111 @@ def _get_coef(self):
...
+                        ast_to_transform))))
+
+    def _power(self, ast_to_transform):
+        power = self.model.model.family.link.power
+        if power == 1:
+            return self._identity(ast_to_transform)
+        elif power == -1:
+            return self._inverse_power(ast_to_transform)
+        elif power == 2:
+            return ast.SqrtExpr(ast_to_transform)
```
izeigerman, line 118: I'm a bit confused here. `SqrtExpr` represents a square root. Why do we use it with power 2 which I assume should mean just a squared expression.

izeigerman: It appears that all powers are "inversed" and it blows my mind :D Can we perhaps inverse it at the very beginning, eg:
```
power = 1 / self.model.model.family.link.power
```
So that the rest of the code makes more sense?
Or rename the `power` variable to something more specific, eg. `inversed_power`.

StrikerRUS: Sorry, I messed up with function names! Should be more clear now after https://github.com/BayesWitnesses/m2cgen/pull/195/commits/0fbaefb9be2a2af6ae5b39d9fbb0a955d8f973bc.
For the reference: https://github.com/statsmodels/statsmodels/blob/master/statsmodels/genmod/families/links.py.

m2cgen/assemblers/linear.py
```diff
@@ -70,6 +75,111 @@ def _get_coef(self):
...
+        elif power == -2:
+            return self._inverse_squared_inversed(ast_to_transform)
+        elif power < 0:  # some languages may not support negative exponent
+            return utils.div(
+                ast.NumVal(1.0),
+                ast.PowExpr(ast_to_transform, ast.NumVal(1 / -power)))
+        else:
+            return ast.PowExpr(ast_to_transform, ast.NumVal(1 / power))
+
+    def _inverse_power_inversed(self, ast_to_transform):
```
izeigerman, line 128: [Nitpick] That's a lot of inversions going on in a single place :) Furthermore they appear to mean somewhat different inversions (negative vs fractional exponent). I don't have any good suggestions here but perhaps you do.

m2cgen/assemblers/linear.py
```diff
@@ -70,6 +75,111 @@ def _get_coef(self):
...
+        return utils.div(ast.NumVal(1.0), ast_to_transform)
+
+    def _sqrt_inversed(self, ast_to_transform):
+        return ast.PowExpr(ast_to_transform, ast.NumVal(2))
+
+    def _inverse_squared_inversed(self, ast_to_transform):
```
izeigerman, line 134: Ditto

izeigerman:
Oops, sorry. I've introduced a conflict by merging your other PR.

m2cgen/assemblers/linear.py
StrikerRUS: Yeah, I don't really like the current names too. The naming schema I used is `_` + `original_name` + `_inversed`. You are absolutely right that `inverse` prefix in original name means "negative sign of argument" while `inversed` postfix corresponds to "opposite" function (e.g. `log <-> exp`). I took `inversed` because we're actually implementing `inverse()` method here of original class.

Wikipedia suggests `anti` suffix: https://en.wikipedia.org/wiki/Inverse_function. But honestly, I don't like that. Maybe `reversed`? To distinguish from original `inverse`. But I guess it is not commonly used.

izeigerman: I don't mind using anti here, but regardless of our choice the meaning will still be somewhat obscure. In any way I don't think that this decision should hold up this PR. I suggest we merge it and decide later. What do you think?

StrikerRUS: Sure! Agree with your plan. 🙂

izeigerman:
Awesome, thank you 👍

StrikerRUS:
For the record.

Support for `probit` and `cauchy` link functions will be added after resolving #196.

Issue #202, opened 2020-05-05
In scikit-learn SVC convert One-vs-one decisions to One-vs-rest


StrikerRUS:
Refer to
https://github.com/BayesWitnesses/m2cgen/blob/d73048ef7d284aa240254324154e7fe18d1571a0/m2cgen/assemblers/svm.py#L110-L127

PR #203, opened 2020-05-05
remove TODO comment

StrikerRUS:
I believe Issues is a better place for `FIXME` and `TODO` things 🙂 .

Just created #202.


izeigerman:
Agree 100%. Let's follow this approach moving forward. Thanks 👍

PR #206, opened 2020-05-07
added Ruby code generator

StrikerRUS:
Ruby is in top15 TIOBE latest rating: https://www.tiobe.com/tiobe-index/. 🙂 

image

I expect that adding Ruby support will be demanded not only in pure Ruby itself, but in **Ruby on Rails** web framework which is among popular frameworks: [1](https://medium.com/front-end-weekly/10-most-popular-web-frameworks-in-2020-167b9103e08a), [2](https://hotframeworks.com/).

### DIFF
--- a/m2cgen/interpreters/ruby/code_generator.py
+++ b/m2cgen/interpreters/ruby/code_generator.py
@@ -1,0 +1,37 @@
+import contextlib
+
+from m2cgen.interpreters.code_generator import ImperativeCodeGenerator
+from m2cgen.interpreters.code_generator import CodeTemplate as CT
+
+
+class RubyCodeGenerator(ImperativeCodeGenerator):
+
+    tpl_var_declaration = CT("")
+    tpl_num_value = CT("${value}")
+    tpl_infix_expression = CT("(${left}) ${op} (${right})")
+    tpl_return_statement = tpl_num_value
+    tpl_array_index_access = CT("${array_name}[${index}]")
+    tpl_if_statement = CT("if ${if_def}")
+    tpl_else_statement = CT("else")
+    tpl_block_termination = CT("end")
+    tpl_var_assignment = CT("${var_name} = ${value}")
+
+    def add_function_def(self, name, args):
+        func_def = "def " + name + "("
+        func_def += ", ".join(args)
+        func_def += ")"
+        self.add_code_line(func_def)
+        self.increase_indent()
+
+    @contextlib.contextmanager
+    def function_definition(self, name, args):
+        self.add_function_def(name, args)
+        yield
+        self.add_block_termination()
+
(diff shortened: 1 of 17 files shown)


m2cgen/interpreters/ruby/interpreter.py
```diff
@@ -0,0 +1,61 @@
...
+        if self.with_linear_algebra:
+            filename = os.path.join(
+                os.path.dirname(__file__), "linear_algebra.rb")
+            self._cg.prepend_code_lines(utils.get_file_content(filename))
+
+        return self._cg.code
+
+    def interpret_bin_num_expr(self, expr, **kwargs):
+        if expr.op == ast.BinNumOpType.DIV:
+            # Always force float result
```
izeigerman, line 49: A brief explanation why would be nice

m2cgen/interpreters/ruby/code_generator.py
```diff
@@ -0,0 +1,37 @@
...
+        self.increase_indent()
+
+    @contextlib.contextmanager
+    def function_definition(self, name, args):
+        self.add_function_def(name, args)
+        yield
+        self.add_block_termination()
+
+    def method_invocation(self, method_name, obj, args):
+        return ("(" + str(obj) + ")." + method_name +
```
izeigerman, line 33: Why do we need these parentheses (the ones in code not in string)? It makes the code a bit harder to digest

izeigerman: Btw, while I'm on it, why do we need parentheses around the object?

m2cgen/interpreters/ruby/interpreter.py
StrikerRUS: image

https://stackoverflow.com/questions/5502761/why-is-division-in-ruby-returning-an-integer-instead-of-decimal-value

m2cgen/interpreters/ruby/code_generator.py
StrikerRUS: It was my initial mistake to not use these parentheses 😃 . It was cough by one of our e2e tests. Then I realized that
`1.0 + 2.0.fdiv(3)` is not the same as `(1.0 + 2.0).fdiv(3)`.

m2cgen/interpreters/ruby/interpreter.py
StrikerRUS: I find using `fdiv` less polluting code than explicit converting every number to float with `.to_f` method.

m2cgen/interpreters/ruby/code_generator.py
izeigerman: ah, makes perfect sense 👍

PR #199, opened 2020-04-30
added fallback expression for Tanh function

StrikerRUS:
Contributes to #196.

Now new languages may not implement or import `Tanh` function, but rely on our fallback expression (but it is **NOT RECOMMENDED**).

All current languages were tested with increased test data fraction to `0.2`: https://travis-ci.org/github/BayesWitnesses/m2cgen/builds/681243574.

Also, here is a proof that new fallback expression works fine:
comment this line
https://github.com/BayesWitnesses/m2cgen/blob/f1caa2055fd07e4edd449f9011ddf44e4c24adac/m2cgen/interpreters/python/interpreter.py#L16
and run

```python
from math import tanh

import numpy as np
import m2cgen as m2c

interpreter = m2c.interpreters.PythonInterpreter()

for i in np.arange(-50, 50, 0.001):
    tanh_expr = m2c.ast.TanhExpr(m2c.ast.NumVal(i))
    exec(interpreter.interpret(tanh_expr))
    np.testing.assert_allclose(tanh(i), score([]), rtol=1e-10)
```

Below are examples of generated code for Java and Haskell when `Tanh` function was not provided

```python
from sklearn.datasets import load_boston
from sklearn.svm import SVR

import m2cgen as m2c

X, y = load_boston(True)
X = X[:10, :4]
y = y[:10]

est = SVR(kernel='sigmoid').fit(X, y)

print(m2c.export_to_java(est))
```

```java
public class Model {

    public static double score(double[] input) {
        double var0;
        double var1;
        var1 = ((0.010678114199077144) * (((((0.00632) * (input[0])) + ((18.0) * (input[1]))) + ((2.31) * (input[2]))) + ((0.0) * (input[3])))) + (0.0);
        if ((var1) > (44.0)) {
            var0 = 1.0;
        } else {
            if ((var1) < (-44.0)) {
                var0 = -1.0;
            } else {
                var0 = (1.0) - ((2.0) / ((Math.exp((2.0) * (var1))) + (1.0)));
            }
        }
        double var2;
        double var3;
        var3 = ((0.010678114199077144) * (((((0.02731) * (input[0])) + ((0.0) * (input[1]))) + ((7.07) * (input[2]))) + ((0.0) * (input[3])))) + (0.0);
        if ((var3) > (44.0)) {
            var2 = 1.0;
        } else {
            if ((var3) < (-44.0)) {
                var2 = -1.0;
            } else {
[...]

PR #208, opened 2020-05-08
added IdExpr

StrikerRUS:
Introduce helper `IdExpr` identity expression, which can be useful in case when you need to overwrite behavior (e.g. `to_reuse` flag) of passed expression.

Refer to https://github.com/BayesWitnesses/m2cgen/pull/199#discussion_r421635242.

### DIFF
--- a/m2cgen/ast.py
+++ b/m2cgen/ast.py
@@ -11,6 +11,17 @@ class Expr:
     to_reuse = False
 
 
+class IdExpr(Expr):
+    def __init__(self, expr, to_reuse=False):
+        self.expr = expr
+        self.to_reuse = to_reuse
+        self.output_size = expr.output_size
+
+    def __str__(self):
+        args = ",".join([str(self.expr), "to_reuse=" + str(self.to_reuse)])
+        return "IdExpr(" + args + ")"
+
+
 class FeatureRef(Expr):
     def __init__(self, index):
         self.index = index
@@ -129,7 +140,7 @@ class VectorExpr(Expr):
 class VectorVal(VectorExpr):
 
     def __init__(self, exprs):
-        assert all(map(lambda e: e.output_size == 1, exprs)), (
+        assert all(e.output_size == 1 for e in exprs), (
             "All expressions for VectorVal must be scalar")
 
         self.exprs = exprs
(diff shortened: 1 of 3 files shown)


m2cgen/ast.py
```diff
@@ -235,7 +246,7 @@ def __str__(self):
     (PowExpr, lambda e: [e.base_expr, e.exp_expr]),
     (VectorVal, lambda e: e.exprs),
     (IfExpr, lambda e: [e.test, e.body, e.orelse]),
-    ((ExpExpr, SqrtExpr, TanhExpr), lambda e: [e.expr]),
+    ((IdExpr, ExpExpr, SqrtExpr, TanhExpr), lambda e: [e.expr]),
```
StrikerRUS, line 249: Not sure here: should it be completely transparent and shouldn't be counted here?..

To be honest, I'd better delegate this part to you @izeigerman because to my shame I still cannot say that I clearly understand how `SubroutinesMixin` works 🙁 .

izeigerman: > I still cannot say that I clearly understand how SubroutinesMixin works

No worries, neither do I 😄 Unfortunately this part of the project is far from being pretty and it's basically an obscure heuristics. I hope I'll have better ideas in future or someone smarter than I will suggest better approaches. 

> should it be completely transparent and shouldn't be counted here?..
I'd say It should be completely transparent and excluded explicitly here https://github.com/BayesWitnesses/m2cgen/blob/9e5af793e8b389d0c55469df938e8ac3b3ccf90d/m2cgen/ast.py#L258

izeigerman: On a second thought I realized that the function is called `count_exprs` so silently ignoring an expression is not the best practice. I'd say we should do one of the following:
1. Do nothing. I don't anticipate this expression to contribute into the count significantly. At least right now.
2. Explicitly exclude using the `exclude_list` argument in places where this function is used.

StrikerRUS: I'm out! Please feel free to choose any variant you think fits here best in your vision of conceptual.

izeigerman: I'd go with the option 2 then - explicitly excluding it in places of invocation.

StrikerRUS:
> Let's exclude the IdExpr explicitly in places where `count_exprs` is used and this is good to go. Thanks!

Sure! As I said earlier, I'm OK with any workaround.

It seems that we had some misunderstanding. Do you want **me** to implement this or **you**'ll add it on your own?

izeigerman:
Ah, sorry about the confusion. We can do either way but I figured since you're on it already... :D Just kidding, I'll add it later

StrikerRUS:
Oops, I'm sorry!

I've just come to the thought that `count_exprs` will not work properly with fallback expressions. Expressions are counted in `_pre_interpret_hook`, but later they are replaced with sequence of other expressions. I might be wrong.

So, I think it's better to do some refactoring after merging #199, and add workaround for `IdExpr` altogether.

Issue #209, opened 2020-05-08
Migrate to f-strings


StrikerRUS:
m2cgen's codebase heavily utilizes string formatting and concatenation mechanisms. [`f-strings`](https://www.python.org/dev/peps/pep-0498/) ([brief guide](https://realpython.com/python-f-strings/)) are known to be the fastest method to format string ([1](https://cito.github.io/blog/f-strings/), [2](https://stackoverflow.com/a/48465349)) and they increase code readability a lot.

![image](https://user-images.githubusercontent.com/25141164/81418871-07a9b980-9156-11ea-8870-6f3bb637789d.png)
Source: https://cito.github.io/blog/f-strings/.

One problem is only that they are supported starting from Python 3.6, but currently m2cgen supports Python 3.5.
https://github.com/BayesWitnesses/m2cgen/blob/7626a601a5f2415c66b4412f6936ed9d87deb77c/setup.py#L28

My suggestion is the following. Mark the next `0.8.0` release as the latest which supports Python 3.5 and drop the support in `0.9.0` release. It will be approximately in the same time as Python 3.5 reaches its EOL ([2020-09-13](https://devguide.python.org/#branchstatus)).

StrikerRUS:
OK, as we've agreed to drop Python 3.5 after 0.8.0 release, I'd like to list all files in the project that should be checked for replacing `format` to f-string. It will be useful because I don't think that one huge PR is a good idea. Instead, a bunch of small PRs is better in terms of review speed and ease.


- [ ] m2cgen\setup.py
- [ ] m2cgen\generated_code_examples\python\classification\decision_tree.py
- [ ] m2cgen\generated_code_examples\python\classification\lightgbm.py
- [ ] m2cgen\generated_code_examples\python\classification\linear.py
- [ ] m2cgen\generated_code_examples\python\classification\random_forest.py
- [ ] m2cgen\generated_code_examples\python\classification\svm.py
- [ ] m2cgen\generated_code_examples\python\classification\xgboost.py
- [ ] m2cgen\generated_code_examples\python\regression\decision_tree.py
- [ ] m2cgen\generated_code_examples\python\regression\lightgbm.py
- [ ] m2cgen\generated_code_examples\python\regression\linear.py
- [ ] m2cgen\generated_code_examples\python\regression\random_forest.py
- [ ] m2cgen\generated_code_examples\python\regression\svm.py
- [ ] m2cgen\generated_code_examples\python\regression\xgboost.py
[...]

PR #213, opened 2020-05-10
migrate from string.Template to string.format

StrikerRUS:
In contrast to #209, this improvement can be done right now before dropping support of Python 3.5.

As it is known that performance highly depends on a kind of templates, I made simple bench for our current `CodeTemplate` class and a new one. And it seems that for our case (very small strings with few placeholders) `format` works best.

> This comparison is, of course, highly dependent on what you're doing. What if you're rendering lots of small templates rather than one massive table? Then entirely different performance characteristics of the template engine would become relevant, like template parsing and loading time. Moral? Make optimization decisions based on your own benchmarks of your own code.
  https://stackoverflow.com/questions/1324238/what-is-the-fastest-template-system-for-python#comment5776737_1326219

> From the jinja2 docs, it seems that string.Template is the fastest if that's all you need.
>>  Without a doubt you should try to remove as much logic from templates as possible. But templates without any logic mean that you have to do all the processing in the code which is boring and stupid. A template engine that does that is shipped with Python and called string.Template. Comes without loops and if conditions and is by far the fastest template engine you can get for Python.

> https://stackoverflow.com/a/1324258


StrikerRUS:
For the bench I collected all possible templates from all currently supported languages.

```
from string import Template


class CodeTemplate:

    def __init__(self, template):
        self.template = Template(template)
        self.str_template = template

    def __str__(self):
        return self.str_template

    def __call__(self, *args, **kwargs):
        return self.template.substitute(*args, **kwargs)


class NewCodeTemplate:

    def __init__(self, template):
        self.str_template = template

    def __str__(self):
        return self.str_template

    def __call__(self, *args, **kwargs):
        # Force calling str() representation
        # because without it numpy gives the same output
        # for different float types
        return self.str_template.format(
            *[str(i) for i in args],
            **{k: str(v) for k, v in kwargs.items()})


values = [1, 42, 42.42, .42, 42.424242424242424242, -2, -42, -42.42, -42.424242424242442, 0]
lefts = values
ops = ["+", "-", "/", "*", "**", "^"]
rights = values
var_types = ["double", "[]double", "Double"]
var_names = ["var0", "var1", "var2"]
array_names = var_names
indexs = [0, 1, 2, 100, 1000]
if_defs = ["(input[5]) > (7.437)", "(input[5]) <= (7.4370003)", "(input[12]) <= (15.0)", "(input[12]) <= (14.395)"]
module_names = ["score", "Model", "Test"]
init_vals = values
type_modifiers = ["", "[]"]
func_names = module_names
block_names = ["Module", "Sub", "Function", "If"]

tpl_1 = "${value}"
tpl_2 = "(${left}) ${op} (${right})"
tpl_3 = "${var_type} ${var_name};"
tpl_4 = "return ${value};"
tpl_5 = "${array_name}[${index}]"
tpl_6 = "if (${if_def}) {{"
tpl_7 = "}} else {{"
tpl_8 = "}}"
tpl_9 = "${var_name} = ${value};"
tpl_10 = "var ${var_name} ${var_type}"
tpl_11 = "return ${value}"
tpl_12 = "if ${if_def} {{"
tpl_13 = "${var_name} = ${value}"
tpl_14 = "module ${module_name} where"
tpl_15 = "$$${array_name}[${index}]"
tpl_16 = "${var_type}${var_name} = ${init_val}"
tpl_17 = "return ${value}"
tpl_18 = "np.asarray(${value})"
tpl_19 = "if ${if_def}:"
tpl_20 = "else:"
tpl_21 = "return(${value})"
tpl_22 = "${var_name} <- ${value}"
tpl_23 = "Dim ${var_name}${type_modifier} As ${var_type}"
tpl_24 = "${func_name} = ${value}"
tpl_25 = "If ${if_def} Then"
tpl_26 = "Else"
tpl_27 = "End ${block_name}"
tpl_28 = "${array_name}(${index})"
tpl_29 = "${array_name}(${index}) = ${value}"
tpl_30 = "Module ${module_name}"

def test():
    class TestCodeTemplateClass(NewCodeTemplate):
        pass
    
    for value in values:
        TestCodeTemplateClass(tpl_1)(value=value)
    
    for left in lefts:
        for op in ops:
            for right in rights:
                TestCodeTemplateClass(tpl_2)(left=left, op=op, right=right)
        
    for var_type in var_types:
        for var_name in var_names:
            TestCodeTemplateClass(tpl_3)(var_type=var_type, var_name=var_name)
                
    for value in values:
        TestCodeTemplateClass(tpl_4)(value=value)
        
    for array_name in array_names:
        for index in indexs:
            TestCodeTemplateClass(tpl_5)(array_name=array_name, index=index)
            
    for if_def in if_defs:
        TestCodeTemplateClass(tpl_6)(if_def=if_def)
        
    for _ in range(3):
        TestCodeTemplateClass(tpl_7)()
        
    for _ in range(3):
        TestCodeTemplateClass(tpl_8)()
        
    for value in values:
        for var_name in var_names:
            TestCodeTemplateClass(tpl_9)(value=value, var_name=var_name)
            
    for var_type in var_types:
        for var_name in var_names:
            TestCodeTemplateClass(tpl_10)(var_type=var_type, var_name=var_name)
            
    for value in values:
        TestCodeTemplateClass(tpl_11)(value=value)
        
    for if_def in if_defs:
        TestCodeTemplateClass(tpl_12)(if_def=if_def)
        
    for value in values:
        for var_name in var_names:
            TestCodeTemplateClass(tpl_13)(value=value, var_name=var_name)
            
    for module_name in module_names:
        TestCodeTemplateClass(tpl_14)(module_name=module_name)
        
    for array_name in array_names:
        for index in indexs:
            TestCodeTemplateClass(tpl_15)(array_name=array_name, index=index)
            
    for var_type in var_types:
        for var_name in var_names:
            for init_val in init_vals:
                TestCodeTemplateClass(tpl_16)(var_type=var_type, var_name=var_name, init_val=init_val)
            
    for value in values:
        TestCodeTemplateClass(tpl_17)(value=value)
        
    for value in values:
        TestCodeTemplateClass(tpl_18)(value=value)
        
    for if_def in if_defs:
        TestCodeTemplateClass(tpl_19)(if_def=if_def)
        
    for _ in range(3):
        TestCodeTemplateClass(tpl_20)()
        
    for value in values:
        TestCodeTemplateClass(tpl_21)(value=value)
        
    for var_name in var_names:
        for value in values:
            TestCodeTemplateClass(tpl_22)(var_name=var_name, value=value)
            
    for var_type in var_types:
        for var_name in var_names:
            for type_modifier in type_modifiers:
                TestCodeTemplateClass(tpl_23)(var_type=var_type, var_name=var_name, type_modifier=type_modifier)
                
    for value in values:
        for func_name in func_names:
            TestCodeTemplateClass(tpl_24)(value=value, func_name=func_name)
            
    for if_def in if_defs:
        TestCodeTemplateClass(tpl_25)(if_def=if_def)
        
    for _ in range(3):
        TestCodeTemplateClass(tpl_26)()
        
    for block_name in block_names:
        TestCodeTemplateClass(tpl_27)(block_name=block_name)
        
    for array_name in array_names:
        for index in indexs:
            TestCodeTemplateClass(tpl_28)(array_name=array_name, index=index)
            
    for array_name in array_names:
        for index in indexs:
            for value in values:
                TestCodeTemplateClass(tpl_29)(array_name=array_name, index=index, value=value)
                
    for module_name in module_names:
        TestCodeTemplateClass(tpl_30)(module_name=module_name)

%%timeit -n30 -r10
test()
```

`CodeTemplate`:

```
18.4 ms ± 1.02 ms per loop (mean ± std. dev. of 10 runs, 30 loops each)
```

`NewCodeTemplate`:

```
9.65 ms ± 1.21 ms per loop (mean ± std. dev. of 10 runs, 30 loops each)
```

StrikerRUS:
Also, I tried the following possible solution, but it gave the worst score.
```
from string import Formatter


class _Formatter(Formatter):
    def convert_field(self, value, conversion):
        # Force calling str() representation by default,
        # because without it numpy gives the same output
        # for different float types
        return super().convert_field(
            value=value,
            conversion="s")


class AnotherCodeTemplate:

    fmt = _Formatter()

    def __init__(self, template):
        self.str_template = template

    def __str__(self):
        return self.str_template

    def __call__(self, *args, **kwargs):
        return self.fmt.format(self.str_template, *args, **kwargs)
```

```
34.4 ms ± 661 µs per loop (mean ± std. dev. of 10 runs, 30 loops each)
```

m2cgen/interpreters/code_generator.py
```diff
@@ -1,17 +1,18 @@
...
         return self.str_template
 
     def __call__(self, *args, **kwargs):
-        return self.template.substitute(*args, **kwargs)
+        # Force calling str() representation
+        # because without it numpy gives the same output
+        # for different float types
+        return self.str_template.format(
+            *[str(i) for i in args],
+            **{k: str(v) for k, v in kwargs.items()})
```
StrikerRUS, line 15: This bug(?) was spotted by our XGBoost(hist) e2e test and can be demonstrated in the following example.

```
import numpy as np

test_float_val = 42.42424242424242424242

print(np.float16(test_float_val).__str__())
# 42.44  Correct
print(np.float32(test_float_val).__str__())
# 42.424244  Correct
print(np.float64(test_float_val).__str__())
# 42.42424242424242  Correct

print(np.float16(test_float_val).__format__(""))
# 42.4375  Wrong
print(np.float32(test_float_val).__format__(""))
# 42.42424392700195  Wrong
print(np.float64(test_float_val).__format__(""))
# 42.42424242424242  Correct
```

izeigerman: Oh wow, this is a great catch. Does it mean that the root cause we originally discovered could've been wrong?

StrikerRUS: Sorry, didn't get it!

This issue illustrates only that we should always dump float values into native code via `str()` and not `format()`. 

Our current code generator use `template.substitute()` which calls `str()` under the hood. So we wasn't able to spot this "bug" (I'm still not sure that this is a real bug, or `numpy` produces different results with `str()` and `format()` by design).

PR #217, opened 2020-05-14
added support for multiclass XGBoost RF

### LINKED ISSUE #163: Add support for XGBoost Random Forest for multiclass task
Initial support for regression and binary classification tasks was added in #157. Unfortunately, multiclass case is not so trivial and requires deeper knowledge of XGBoost internals (dumped model representation and prediction logic) to add support for it.

StrikerRUS:
Closed #163.

Found solution by brute forcing all variants for tiny model 😄 
```
est = xgb.XGBRFClassifier(base_score=0, n_estimators=2, max_depth=1).fit(X, y)
```


m2cgen/assemblers/boosting.py
```diff
@@ -10,6 +10,7 @@
 class BaseBoostingAssembler(ModelAssembler):
 
     classifier_names = {}
+    strided_layout_for_multiclass = True
```
izeigerman, line 13: Forgive me my ignorance but what exactly `strided` mean in this context? :)

StrikerRUS: You shouldn't apologize! I was really struggling with the good name for this parameter! 😄 
 Initially I thought about `step` word in the name. Then `stride` came to my mind from convolutional nets. I started to google for any synonyms and came to the following picture

image
(taken from 3rd slide of https://slideplayer.com/slide/9359236/)

izeigerman: Got it, thank you!

PR #220, opened 2020-05-18
added fallback expressions for Sqrt and Exp functions

StrikerRUS:
Continuation of #199.


m2cgen/assemblers/fallback_expressions.py
```diff
@@ -30,6 +30,20 @@ def tanh(expr):
...
+def sqrt(expr, to_reuse=False):
+    return ast.PowExpr(
+        base_expr=expr,
+        exp_expr=ast.NumVal(0.5),
+        to_reuse=to_reuse)
+
+
+def exp(expr, to_reuse=False):
+    return ast.PowExpr(
+        base_expr=ast.NumVal(2.71828182845904523536028747135),
```
izeigerman, line 42: @StrikerRUS Sorry for commenting in the closed PR but I've just realized - isn't there a constant in Python standard library for this?

StrikerRUS: Indeed, my mistake! I'll fix it right now.

I'm too deep in fallback expressions right now that I'm afraid to use anything already implemented 😃 .

PR #221, opened 2020-05-20
support not only task-default objective functions in LightGBM

StrikerRUS:
Refer to https://lightgbm.readthedocs.io/en/latest/Parameters.html#objective and https://github.com/microsoft/LightGBM/tree/master/src/objective.

Also, contributes to #99 in terms of listing all supported objectives per task.

TODO:
- [x] `abs` function (#222)
- [x] `log1p` function (#225)


StrikerRUS:
I think it is ready for review.

m2cgen/assemblers/boosting.py
```diff
@@ -225,6 +237,76 @@ def _final_transform(self, ast_to_transform):
...
+        return supported_objectives[self.objective_name](expr)
+
+    def _log1p_exp_transform(self, expr):
+        return ast.Log1pExpr(ast.ExpExpr(expr))
+
+    def _maybe_sqr_transform(self, expr):
+        need_sqr = "sqrt" in self.objective_config.split(" ")
+        if need_sqr:
+            expr = ast.IdExpr(expr, to_reuse=True)
+            return utils.mul(ast.AbsExpr(expr), expr)
```
izeigerman, line 303: I don't quite understand the purpose of the Abs here. Can you please elaborate this expression

m2cgen/assemblers/boosting.py
```diff
@@ -225,6 +237,76 @@ def _final_transform(self, ast_to_transform):
...
+        if self.objective_name not in supported_objectives:
+            raise ValueError(
+                f"Unsupported objective function '{self.objective_name}'")
+        return supported_objectives[self.objective_name](expr)
+
+    def _log1p_exp_transform(self, expr):
+        return ast.Log1pExpr(ast.ExpExpr(expr))
+
+    def _maybe_sqr_transform(self, expr):
+        need_sqr = "sqrt" in self.objective_config.split(" ")
```
izeigerman, line 300: It seems like we do splitting quite a few times does it make sense to do this once in the constructor and have an attribute like `objective_config_splits`.

StrikerRUS: Ah, good idea! Making a commit right now.

StrikerRUS: Done in 1defd1f39a82dc96c9c0946a6f319389baaf4e39.

m2cgen/assemblers/boosting.py
StrikerRUS: Sure! We should preserve a sign of original result according to the original code. I decided to go with `abs` expression (instead of `sign`) because I think it is more common and may be used in future assemblers.

> `output[0] = Common::Sign(input[0]) * input[0] * input[0];`

https://github.com/microsoft/LightGBM/blob/4f8c32d9a6c6b0c8d774d571da3bcd70191d218b/src/objective/regression_objective.hpp#L146-L152

izeigerman: Makes sense, thank you!

PR #222, opened 2020-05-20
added Abs function

StrikerRUS:
`abs` function is needed for #221.

I extracted this part into separate PR to not overcomplicate original one.


m2cgen/interpreters/python/interpreter.py
```diff
@@ -39,6 +40,11 @@ def interpret(self, expr):
 
         return self._cg.finalize_and_get_generated_code()
 
+    def interpret_abs_expr(self, expr, **kwargs):
+        result = super().interpret_abs_expr(expr, **kwargs)
+        self.with_math_module = False
```
izeigerman, line 45: I'm not sure setting this flag back to false here is a good idea. We can't really guarantee the order of interpretation of expressions and if for some reason this one is being interpreted after all other math functions - we'll get a missing import.

StrikerRUS: Ah, you're absolutely right!
I added this line to not add `import math` for `abs`, but indeed it can break previous interpretations. 🤦‍♂️ 

One way to fix it is to write `interpret_abs_expr` from scratch without `self.with_math_module` but it duplicates the code. Don't you have any better ideas in mind?

izeigerman: Duplicating part of the logic doesn't sound like a too bad of an idea considering that only these 2 lines are required:
```
        nested_result = self._do_interpret(expr.expr, **kwargs)
        return self._cg.function_invocation(
            self.abs_function_name, nested_result)
```
I assume we don't care about all the fallback logic here.

StrikerRUS: > I assume we don't care about all the fallback logic here.

Makes sense!

OK, I'll follow your proposal. Thank you!

StrikerRUS: Done!

StrikerRUS: OK, now we cannot use `PythonInterpreter` in `test_abs_fallback_expr` due to we removed the fallback logic from there. Changing it to `CInterpreter`.

PR #225, opened 2020-05-26
added log and log1p functions

StrikerRUS:
`log1p` function is needed for #221.

I extracted this part into separate PR to not overcomplicate original one.

Custom implementations of `log1p` are based on 

https://github.com/statgen/topmed_freeze3_calling/blob/master/vt/lib/Rmath/chebyshev.c
https://hackage.haskell.org/package/statistics-0.8.0.0/docs/src/Statistics-Math.html


m2cgen/interpreters/interpreter.py
```diff
@@ -130,6 +132,23 @@ def interpret_exp_expr(self, expr, **kwargs):
         return self._cg.function_invocation(
             self.exponent_function_name, nested_result)
 
+    def interpret_log_expr(self, expr, **kwargs):
+        if self.logarithm_function_name is NotImplemented:
+            raise NotImplementedError("Logarithm function is not provided")
```
StrikerRUS, line 137: Unfortunately, I didn't find anything that we could use as a fallback expression. All methods are based on **reduction** technique  which requires `frexp` implementation.
Refer for example to https://golang.org/src/math/log1p.go

m2cgen/interpreters/interpreter.py
```diff
@@ -140,6 +142,23 @@ def interpret_exp_expr(self, expr, **kwargs):
...
+        self.with_math_module = True
+        nested_result = self._do_interpret(expr.expr, **kwargs)
+        return self._cg.function_invocation(
+            self.logarithm_function_name, nested_result)
+
+    def interpret_log1p_expr(self, expr, **kwargs):
+        self.with_math_module = True
```
izeigerman, line 154: I believe we should set this flag only in case when we don't follow the fallback path. In case of the fallback we technically don't require the math module, although subsequent expressions that constitute the fallback expression may choose to enable it.

StrikerRUS: > I can't believe you ported the log1p function to all those languages

Thanks! Actually it was not so hard.  All imperative languages are so similar! 😄 

> I believe we should set this flag only in case when we don't follow the fallback path.

Sorry, I'm not sure I fully understood you. Do you mean the following?
```
if self.log1p_function_name is NotImplemented:
            return self._do_interpret(
                fallback_expressions.log1p(expr.expr), **kwargs)
self.with_math_module = True
```

If so, it seems reasonable for me. But we should double check cases when fallbacks use functions from math module.

Anyway, I believe that this should be addressed in a separate PR because all other expressions follow the same pattern for now.

izeigerman: Yeah, this is exactly what I mean!

> But we should double check cases when fallbacks use functions from math module.

Yep, but those functions will also be interpreted and the flag will be set accordingly if needed.

 > Anyway, I believe that this should be addressed in a separate PR because all other expressions follow the same pattern for now.

I'm fine with that. We may want to address this for other functions as well.

StrikerRUS: Thanks! I created new issue for that to not forget: #246.

izeigerman:
@StrikerRUS I believe we now have to add `__eq__` and `__hash__` for new expressions before merging this PR, right?

StrikerRUS:
Thanks a lot for the reminder! Done in a6d20464c0fc9dc3baed1e243d32a47f7b19e0fa.

PR #216, opened 2020-05-14
added sklearn GLM

StrikerRUS:
Closed #214.

PR #227, opened 2020-06-01
Drop Python 3.5 support

StrikerRUS:
Reasons:
 - #209 
   - f-strings
- https://github.com/BayesWitnesses/m2cgen/pull/216#discussion_r424819545
  - new scikit-learn version is incompatible with Python 3.5
  - new XGBoost version is incompatible with Python 3.5


izeigerman:
Not merging this until we release 0.8.0

PR #243, opened 2020-06-12
setup GitHub Actions

StrikerRUS:
Hmm, seems that Actions can be triggered even without config in the `master` branch:

image

One more step is to make passing tests at GitHub Actions for PRs **required**. Refer to https://github.com/microsoft/LightGBM/pull/3119#discussion_r432846447


.github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+          fetch-depth: 5
+      - name: Setup Python ${{ matrix.python }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python }}
+      - name: Setup dependencies and run tests
+        run: |
+          if [[ "${{ matrix.lang }}" == "API" ]]; then
+              export TEST="API";
+              export COVERALLS_REPO_TOKEN="COVERALLS TOKEN FOR M2CGEN";
```
StrikerRUS, line 41: FIXME

.github/workflows/main.yml
```diff
@@ -0,0 +1,54 @@
...
+          else
+              export TEST="E2E";
+              export LC_ALL="en_US.UTF-8";
+              sudo locale-gen $LC_ALL;
+              sudo update-locale;
```
StrikerRUS, line 46: Workaround for
```
process terminated. infinite recursion during resource lookup within system.private.corelib.
```
https://github.com/dotnet/runtime/issues/32510

.github/workflows/main.yml
izeigerman: I'm not quite sure how to overcome this. It doesn't seem to be a good idea to expose the API token and make it a part of the publicly available source code. Or am I missing something?

izeigerman: Shall we use a Coveralls action (https://github.com/marketplace/actions/coveralls-github-action) instead of manually invoking the `coveralls` binary in the `test.sh` script?

StrikerRUS:
Oh, I just noticed that the Action with R language takes more than 2 hours (all other Actions take approximately the same time as they do at Travis)! It looks like either Travis has better CPUs or Rscript conflicts with some installed software in GitHub Actions image.

StrikerRUS:
I'm pretty sure we should use Coveralls Action instead of manual invocation in bash script. But unfortunately, I have no experience with it.

Speaking about the Travis, the only reason I see to keep it is even more parallel jobs (possibly later with greater number of supported languages and models). Maybe we can run `r_lang` job at Travis as it takes much less time there? Or it is overkill?..

izeigerman:
> Maybe we can run r_lang job at Travis as it takes much less time there? Or it is overkill?.

That's actually a very interesting idea, tough I'd rather prefer to have a consistent and uniform environment for all our tests. We should keep this in mind.

> I'm pretty sure we should use Coveralls Action instead of manual invocation in bash script.

I'll take a look into this one

StrikerRUS:
So, we are removing Travis config for now (after migrating to Coveralls Action), right?

Also, please don't forget to make GitHub Actions checks required: https://github.com/BayesWitnesses/m2cgen/pull/243#issue-433916764.

One more frustrating thing is that each time config (matrix or jobs name) changes, you need to go to repo settings and update the list of required checks. Refer to https://github.com/microsoft/LightGBM/pull/3188#issuecomment-657270592 (paragraph starting with `GitHub Actions uses ...`). For example, #262 changes the set of jobs in matrix which means you'll need to update required checks.

izeigerman:
> So, we are removing Travis config for now (after migrating to Coveralls Action), right?

Yeah, let's do that.

StrikerRUS:
> tough I'd rather prefer to have a consistent and uniform environment for all our tests.

BTW, we can use [our Docker](https://github.com/BayesWitnesses/m2cgen/blob/master/Dockerfile) file for that purpose.

PR #253, opened 2020-06-23
added test for version CLI argument

StrikerRUS:


tests/test_cli.py
```diff
@@ -74,6 +74,16 @@ def test_language_is_required(mocked_exit):
     mocked_exit.assert_called_with(2)
 
 
+@mock.patch.object(sys, "exit")
+def test_version(mocked_exit):
```
izeigerman, line 78: I'm not sure why do we need the `mocked_exit` fixture here though. It doesn't seem to be used.

StrikerRUS: Yeah, I was thinking so too! But without this mock test fails with 
```
self = ArgumentParser(prog='m2cgen', usage=None, description='Generate code in native language for provided model.', formatter_class=<class 'argparse.HelpFormatter'>, conflict_handler='error', add_help=True)
status = 0, message = None

    def exit(self, status=0, message=None):
        if message:
            self._print_message(message, _sys.stderr)
>       _sys.exit(status)
E       SystemExit: 0

/opt/python/3.6.7/lib/python3.6/argparse.py:2380: SystemExit
```

See my [initial commit](https://github.com/BayesWitnesses/m2cgen/pull/253/commits/d76192cf1c0acf2c1a8f83dfd6f6f41721d72b79) here and the [corresponding CI job](https://travis-ci.org/github/BayesWitnesses/m2cgen/jobs/701120157).

izeigerman: Got it, thanks!

PR #257, opened 2020-06-25
add base_score only when needed

StrikerRUS:
It was quite strange to see global bias in generated code for LightGBM. Also, it simplifies hence speeds up runtime.


m2cgen/assemblers/boosting.py
```diff
@@ -45,9 +45,14 @@ def _assemble_single_output(self, estimator_params,
 
         tmp_ast = utils.apply_op_to_expressions(
             ast.BinNumOpType.ADD,
-            ast.NumVal(base_score),
             *estimators_ast)
 
+        if base_score != 0.0:
```
izeigerman, line 50: This is a great idea 👍 Do you think the same logic can be applied to bias and coefficients in case of linear models? I can see how L1-regularized linear models can benefit from this greatly (relatively).

StrikerRUS: Of course, awesome proposal! But I have plans to introduce `DotProduct` expression with a fallback (very similar one to the current `linear_to_ast` function). I think I will address sparsity problems there.
https://github.com/BayesWitnesses/m2cgen/blob/ec02c727c73f5cb0dc83b13a5514a0bb1304f1ae/m2cgen/assemblers/linear.py#L225-L233

PR #231, opened 2020-06-05
improve caching mechanism

StrikerRUS:
For now we can only reuse expression if it is the same piece of memory:

```
from m2cgen import ast
from m2cgen.interpreters import PythonInterpreter

expr1 = ast.BinNumExpr(ast.NumVal(1), ast.NumVal(1), ast.BinNumOpType.ADD, to_reuse=True)
expr2 = ast.BinNumExpr(ast.NumVal(1), ast.NumVal(1), ast.BinNumOpType.ADD, to_reuse=True)
expr = ast.BinNumExpr(expr1, expr2, ast.BinNumOpType.ADD)

interpreter = PythonInterpreter()

print(interpreter.interpret(expr))
```

```
def score(input):
    var0 = (1) + (1)
    var1 = (1) + (1)
    return (var0) + (var1)
```

For example, look how the amount of generated SVM code decreased with this improvement.

```
from sklearn.datasets import load_digits
from sklearn.svm import SVC
from sklearn.model_selection import train_test_split

import m2cgen as m2c

X, y = load_digits(3, True)
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.2, random_state=42)
X_train = X_train[:15, :3]
y_train = y_train[:15]

est = SVC().fit(X_train, y_train)

print(m2c.export_to_python(est))
```

Before:
```
import math
def score(input):
    var0 = (0) - (0.026918168766948466)
    var1 = math.exp((var0) * (((math.pow((0.0) - (input[0]), 2)) + (math.pow((0.0) - (input[1]), 2))) + (math.pow((1.0) - (input[2]), 2))))
    var2 = math.exp((var0) * (((math.pow((0.0) - (input[0]), 2)) + (math.pow((0.0) - (input[1]), 2))) + (math.pow((0.0) - (input[2]), 2))))
    var3 = math.exp((var0) * (((math.pow((0.0) - (input[0]), 2)) + (math.pow((0.0) - (input[1]), 2))) + (math.pow((1.0) - (input[2]), 2))))
    var4 = math.exp((var0) * (((math.pow((0.0) - (input[0]), 2)) + (math.pow((0.0) - (input[1]), 2))) + (math.pow((0.0) - (input[2]), 2))))
    var5 = math.exp((var0) * (((math.pow((0.0) - (input[0]), 2)) + (math.pow((0.0) - (input[1]), 2))) + (math.pow((3.0) - (input[2]), 2))))
[...]

PR #262, opened 2020-07-01
added F# code generator

StrikerRUS:
Now the whole .NET family are together here!

Linking #127.


m2cgen/ast.py
```diff
@@ -54,10 +56,8 @@ class NumExpr(Expr):
 
 
 class NumVal(NumExpr):
-    def __init__(self, value, dtype=None):
-        if dtype:
-            value = dtype(value)
-        self.value = value
+    def __init__(self, value, dtype=np.float64):
+        self.value = dtype(value)
```
StrikerRUS, line 60: Some languages prohibit implicit type casting like F#. Therefore, we should ensure that each value is double unless we explicitly specify another type. Refer to https://stackoverflow.com/questions/45710437/why-cant-you-add-int-and-float-literals-in-f.

m2cgen/interpreters/f_sharp/interpreter.py
```diff
@@ -0,0 +1,120 @@
...
+        if self._cached_expr_results:
+            for func_name, expr_result in self._cached_expr_results.values():
+                self._cg.add_function(
+                    function_name=func_name, function_body=expr_result)
+
+    def bin_depth_threshold_hook(self, expr, **kwargs):
+        if expr in self._cached_expr_results:
+            return self._cached_expr_results[expr].var_name
+        result = self._do_interpret(expr, **kwargs)
+        return self._cache_reused_expr(expr, result)
```
StrikerRUS, line 120: I think it is OK to use caching in a particular interpreter but not in a base class. Refer to https://github.com/BayesWitnesses/m2cgen/pull/231#discussion_r442295716.

m2cgen/interpreters/f_sharp/interpreter.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if self.with_log1p_expr:
+            filename = os.path.join(
+                os.path.dirname(__file__), "log1p.fs")
+            self._cg.prepend_code_lines(utils.get_file_content(filename))
+
+        return self._cg.finalize_and_get_generated_code()
+
+    def interpret_if_expr(self, expr, if_code_gen=None, **kwargs):
+        if if_code_gen is None:
```
izeigerman, line 69: This begins to look like a common pattern among interpreters for functional languages. We should consider generalizing this at some point.

PR #272, opened 2020-07-22
Add the Coveralls step to the Github Actions workflow

izeigerman:


izeigerman:
Ended up using the coverall-python directly. See the [issue](https://github.com/coverallsapp/github-action/issues/4) for details.

izeigerman:
The runtime for `r_lang` is rather ridiculous. @StrikerRUS I'm leaning towards following your advice and keep it on Travis for now. Although I don't like the idea of using 2 services for the exact same purpose it appears to be a lesser evil than a ~180 minute runtime. What do you think?

StrikerRUS:
I think we can drop Travis CI for now. I have one idea how to fix long R runtime tests. If it doesn't work, we can add Travis config with only R job back. I believe that with current frequency of PRs long CI jobs are not something crucial for us.

.github/workflows/main.yml
```diff
@@ -10,7 +10,36 @@ on:
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          TEST: API
+          PYTHON: ${{ matrix.python }}
+        run: |
+          export BUILD_DIRECTORY=$GITHUB_WORKSPACE
+          bash $GITHUB_WORKSPACE/.ci/setup.sh
+          bash $GITHUB_WORKSPACE/.ci/test.sh
+
+  e2e-test:
+    name: "Python ${{ matrix.python }}, Test: '${{ matrix.lang }}'"
```
StrikerRUS, line 42: We can use this value without escaping. Or were these quotes added just for aesthetic purposes?
```suggestion
    name: "Python ${{ matrix.python }}, Test: ${{ matrix.lang }}"
```

.github/workflows/main.yml
```diff
@@ -33,20 +61,16 @@ jobs:
         uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python }}
-      - name: Setup dependencies and run tests
+      - name: Setup dependencies and run E2E tests
+        env:
+          TEST: E2E
+          LANG: ${{ matrix.lang }}
+          PYTHON: ${{ matrix.python }}
```
StrikerRUS, line 68:
```suggestion
          PYTHON: ${{ matrix.python }}
          BUILD_DIRECTORY: ${{ github.workspace }}
```

.github/workflows/main.yml
```diff
@@ -33,20 +61,16 @@ jobs:
...
-              sudo update-locale;
-          fi
-          export PYTHON="${{ matrix.python }}"
-          export LANG="${{ matrix.lang }}"
+          export LC_ALL="en_US.UTF-8"
+          sudo locale-gen $LC_ALL
+          sudo update-locale
+
           export BUILD_DIRECTORY=$GITHUB_WORKSPACE
           bash $GITHUB_WORKSPACE/.ci/setup.sh
```
StrikerRUS, line 75:
```suggestion
          bash $GITHUB_WORKSPACE/.ci/setup.sh
```

.github/workflows/main.yml
izeigerman: Yeah - the latter :D Reverting it back

izeigerman:
@StrikerRUS thanks for the review! I've addressed your comments and dropped the Travis config.

izeigerman:
@StrikerRUS I've just realized that it's super annoying to maintain the list of required checks using their names since we can reshuffle language E2E tests. Plus github stores the history of all builds:
Screen Shot 2020-07-23 at 9 24 25 AM
The list of checks will keep growing and picking the right check will become rather error prone. I wonder whether we should consider removing the `LANG` variable from the job name. It will look less nice but at least it should be easier to maintain.

PR #273, opened 2020-07-23
introduce FunctionalCodeGenerator and FunctionalToCodeInterpreter

StrikerRUS:
Refer to https://github.com/BayesWitnesses/m2cgen/pull/262#discussion_r458887319.


m2cgen/interpreters/f_sharp/code_generator.py
```diff
@@ -1,41 +1,21 @@
...
-
     def add_if_termination(self):
         self.decrease_indent()
 
-    def get_func_name(self):
-        func_name = f"func{self._func_idx}"
-        self._func_idx += 1
-        return func_name
-
-    def add_function(self, function_name, function_body):
```
izeigerman, line 33: `function_invocation` looks exactly the same for F# and Haskell too

m2cgen/interpreters/haskell/code_generator.py
```diff
@@ -1,48 +1,31 @@
...
-        self._func_idx = 0
-
     def array_index_access(self, array_name, index):
         return self.tpl_infix_expression(
             left=array_name, op="!!", right=index)
 
     def add_if_statement(self, if_def):
-        self.add_code_line(f"if ({if_def})")
-        self.increase_indent()
+        super().add_if_statement(if_def=if_def)
```
izeigerman, line 21: Is it really necessary to move the `then` statement to the next line? I wonder if we should just reuse the parent  method and make `then` a part of the `tpl_if_statement`.

m2cgen/interpreters/haskell/code_generator.py
```diff
@@ -1,48 +1,31 @@
...
     def add_if_termination(self):
         self.decrease_indent()
         self.decrease_indent()
 
-    def get_func_name(self):
-        func_name = f"func{self._func_idx}"
-        self._func_idx += 1
-        return func_name
-
-    def add_function(self, function_name, function_body):
```
izeigerman, line 40: Ditto for the same `function_invocation` implementation.

m2cgen/interpreters/interpreter.py
```diff
@@ -228,3 +228,51 @@ def _cache_reused_expr(self, expr, expr_result):
...
             var_name=var_name, expr_result=None)
         return var_name
+
+
+class FunctionalToCodeInterpreter(ToCodeInterpreter):
+    """
+    This interpreter provides default implementation for the methods
+    interpreting AST expression into code.
+
+    It can be used for the most programming languages and requires only
```
izeigerman, line 238: This paragraph doesn't seem to be accurate for this base class.

m2cgen/interpreters/interpreter.py
```diff
@@ -228,3 +228,51 @@ def _cache_reused_expr(self, expr, expr_result):
         self._cached_expr_results[expr] = CachedResult(
             var_name=var_name, expr_result=None)
         return var_name
+
+
+class FunctionalToCodeInterpreter(ToCodeInterpreter):
```
izeigerman, line 233: `interpret_pow_expr` and `interpret_log1p_expr` also seem to have the exact same implementations for F# and Haskell

m2cgen/interpreters/f_sharp/code_generator.py
StrikerRUS: Done in 464f285c718e5077fea6f9aaaa9db3131867e741.

m2cgen/interpreters/haskell/code_generator.py
StrikerRUS: Done in 464f285c718e5077fea6f9aaaa9db3131867e741.
I didn't want to modify this to make the diff smaller (due to API tests). 😃

m2cgen/interpreters/haskell/code_generator.py
StrikerRUS: Done in 464f285c718e5077fea6f9aaaa9db3131867e741.

m2cgen/interpreters/interpreter.py
StrikerRUS: Done in 464f285c718e5077fea6f9aaaa9db3131867e741.
Added the word `functional`.

m2cgen/interpreters/interpreter.py
StrikerRUS: I don't think that interpreters of particular functions should be in the base class. I believe it should be as much general as possible. There are a lot more functional programming languages than F# and Haskell. For example, Clojure and Erlang have `pow` function.

PR #277, opened 2020-07-26
handle floating point values more accurate

StrikerRUS:
I run all tests with the increased test dataset fraction (`0.6`) and compared results with ones obtained from `master` but with changed dataset splitting routine to ensure that inputs are the same in both cases.

```diff
diff --git a/tests/e2e/test_e2e.py b/tests/e2e/test_e2e.py
index b2aacb5..2644547 100644
--- a/tests/e2e/test_e2e.py
+++ b/tests/e2e/test_e2e.py
@@ -38,7 +38,7 @@ CLASSIFICATION = pytest.mark.clf
 
 
 # Set of helper functions to make parametrization less verbose.
-def regression(model, test_fraction=0.02):
+def regression(model, test_fraction=0.6):
     return (
         model,
         utils.get_regression_model_trainer(test_fraction),
@@ -46,7 +46,7 @@ def regression(model, test_fraction=0.02):
     )
 
 
-def classification(model, test_fraction=0.02):
+def classification(model, test_fraction=0.6):
     return (
         model,
         utils.get_classification_model_trainer(test_fraction),
@@ -54,7 +54,7 @@ def classification(model, test_fraction=0.02):
     )
 
 
-def classification_binary(model, test_fraction=0.02):
+def classification_binary(model, test_fraction=0.6):
     return (
         model,
         utils.get_binary_classification_model_trainer(test_fraction),
@@ -62,7 +62,7 @@ def classification_binary(model, test_fraction=0.02):
     )
 
 
-def regression_random(model, test_fraction=0.02):
+def regression_random(model, test_fraction=0.6):
     return (
         model,
         utils.get_regression_random_data_model_trainer(test_fraction),
@@ -70,7 +70,7 @@ def regression_random(model, test_fraction=0.02):
     )
 
 
-def classification_random(model, test_fraction=0.02):
+def classification_random(model, test_fraction=0.6):
     return (
         model,
         utils.get_classification_random_data_model_trainer(test_fraction),
@@ -78,7 +78,7 @@ def classification_random(model, test_fraction=0.02):
     )
 
 
-def classification_binary_random(model, test_fraction=0.02):
+def classification_binary_random(model, test_fraction=0.6):
     return (
         model,
         utils.get_classification_binary_random_data_model_trainer(
@@ -87,7 +87,7 @@ def classification_binary_random(model, test_fraction=0.02):
     )
 
 
-def regression_bounded(model, test_fraction=0.02):
+def regression_bounded(model, test_fraction=0.6):
     return (
         model,
         utils.get_bounded_regression_model_trainer(test_fraction),
@@ -228,11 +228,11 @@ STATSMODELS_LINEAR_REGULARIZED_PARAMS = dict(method="elastic_net",
 
         # XGBoost (tree method "hist")
         regression(xgboost.XGBRegressor(**XGBOOST_HIST_PARAMS),
-                   test_fraction=0.2),
+                   test_fraction=0.6),
         classification(xgboost.XGBClassifier(**XGBOOST_HIST_PARAMS),
-                       test_fraction=0.2),
+                       test_fraction=0.6),
         classification_binary(xgboost.XGBClassifier(**XGBOOST_HIST_PARAMS),
-                              test_fraction=0.2),
+                              test_fraction=0.6),
 
         # XGBoost (LINEAR)
         regression(xgboost.XGBRegressor(**XGBOOST_PARAMS_LINEAR)),
diff --git a/tests/utils.py b/tests/utils.py
index 46bae2d..f72b1a3 100644
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -14,7 +14,7 @@ from lightning.impl.base import BaseClassifier as LightBaseClassifier
 from sklearn import datasets
 from sklearn.base import BaseEstimator, RegressorMixin, clone
 from sklearn.ensemble._forest import ForestClassifier
-from sklearn.utils import shuffle
+from sklearn.model_selection import train_test_split
 from sklearn.linear_model._base import LinearClassifierMixin
 from sklearn.tree import DecisionTreeClassifier
 from sklearn.svm import SVC, NuSVC
@@ -70,25 +70,17 @@ class ModelTrainer:
         np.random.seed(seed=7)
         if dataset_name == "boston":
             self.name = "train_model_regression"
-            dataset = datasets.load_boston()
-            self.X, self.y = shuffle(
-                dataset.data, dataset.target, random_state=13)
+            self.X, self.y = datasets.load_boston(True)
         elif dataset_name == "boston_y_bounded":
             self.name = "train_model_regression_bounded"
-            dataset = datasets.load_boston()
-            self.X, self.y = shuffle(
-                dataset.data, dataset.target, random_state=13)
+            self.X, self.y = datasets.load_boston(True)
             self.y = np.arctan(self.y) / np.pi + 0.5  # (0; 1)
         elif dataset_name == "iris":
             self.name = "train_model_classification"
-            dataset = datasets.load_iris()
-            self.X, self.y = shuffle(
-                dataset.data, dataset.target, random_state=13)
+            self.X, self.y = datasets.load_iris(True)
         elif dataset_name == "breast_cancer":
             self.name = "train_model_classification_binary"
-            dataset = datasets.load_breast_cancer()
-            self.X, self.y = shuffle(
-                dataset.data, dataset.target, random_state=13)
+            self.X, self.y = datasets.load_breast_cancer(True)
         elif dataset_name == "regression_rnd":
             self.name = "train_model_regression_random_data"
             N = 1000
@@ -107,9 +99,9 @@ class ModelTrainer:
         else:
             raise ValueError("Unknown dataset name: {}".format(dataset_name))
 
-        offset = int(self.X.shape[0] * (1 - test_fraction))
-        self.X_train, self.y_train = self.X[:offset], self.y[:offset]
-        self.X_test, self.y_test = self.X[offset:], self.y[offset:]
+        (self.X_train, self.X_test,
+         self.y_train, self.y_test) = train_test_split(
+            self.X, self.y, test_size=test_fraction, random_state=13)
 
     @classmethod
     def get_instance(cls, dataset_name, test_fraction=0.02):
```

image

Seems that at least it doesn't make things worse, but even decreases the number of failed tests sometimes.


tests/utils.py
```diff
@@ -125,15 +119,22 @@ def __call__(self, estimator):
...
+                 XGBClassifier, LGBMClassifier)):
             y_pred = estimator.predict_proba(self.X_test)
         else:
             y_pred = estimator.predict(self.X_test)
 
+        # Some models force input data to be particular type
+        # during prediction phase in their native Python libraries.
+        # For correct comparison of testing results we mimic the same behavior
+        if isinstance(estimator, (BaseDecisionTree, BaseForest)):
+            self.X_test = self.X_test.astype(np.float32, copy=False)
```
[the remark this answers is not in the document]
StrikerRUS: Interestingly that XGBoost also cast all inputs to `float` during `predict`.

https://github.com/dmlc/xgboost/blob/12110c900eff0aaa06045ecf717e6c5a36a164d5/python-package/xgboost/sklearn.py#L595-L596

https://github.com/dmlc/xgboost/blob/12110c900eff0aaa06045ecf717e6c5a36a164d5/python-package/xgboost/core.py#L433-L438

https://github.com/dmlc/xgboost/blob/12110c900eff0aaa06045ecf717e6c5a36a164d5/python-package/xgboost/data.py#L493-L495

https://github.com/dmlc/xgboost/blob/12110c900eff0aaa06045ecf717e6c5a36a164d5/python-package/xgboost/data.py#L135

https://github.com/dmlc/xgboost/blob/12110c900eff0aaa06045ecf717e6c5a36a164d5/python-package/xgboost/data.py#L115-L116

https://github.com/dmlc/xgboost/blob/12110c900eff0aaa06045ecf717e6c5a36a164d5/python-package/xgboost/data.py#L138

But including it here makes a lot of tests to fail...

tests/utils.py
```diff
@@ -125,15 +119,22 @@ def __call__(self, estimator):
...
+        elif isinstance(estimator, BaseLibSVM):
+            self.X_test = self.X_test.astype(np.float64, copy=False)
```
StrikerRUS, line 136: https://github.com/scikit-learn/scikit-learn/blob/4a60ec129d3088d095b30cf54a670fd596ca4cc8/sklearn/svm/_base.py#L333

https://github.com/scikit-learn/scikit-learn/blob/4a60ec129d3088d095b30cf54a670fd596ca4cc8/sklearn/svm/_base.py#L464-L466

tests/utils.py
```diff
@@ -238,9 +239,9 @@ def predict_from_commandline(exec_args):
     items = stdout.decode("utf-8").strip().split(" ")
 
     if len(items) == 1:
-        return float(items[0])
+        return np.float64(items[0])
     else:
-        return [float(i) for i in items]
+        return [np.float64(i) for i in items]
```
StrikerRUS, line 244: Let's use numpy types across all codebase for the consistency.

izeigerman:
This is some amazing investigation (as always)! I'll take a look soon. Thank you 👍

tests/utils.py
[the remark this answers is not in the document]
izeigerman: I'm a little concerned that here we treat a symptom not the cause. By casting the input vector to float32 and them passing it as strings into estimators we don't exactly reproduce the actual environment, where  casted values will be transformed back to doubles because of the `score` function signature. What do you think? Am I overthinking this?

tests/assemblers/test_xgboost.py
```diff
@@ -352,17 +352,17 @@ def test_regression_random_forest():
...
                     ast.CompOpType.GTE),
-                ast.NumVal(18.1008453),
-                ast.NumVal(9.60167599)),
+                ast.NumVal(17.4066162109),
+                ast.NumVal(9.6789960861)),
             ast.IfExpr(
                 ast.CompExpr(
-                    ast.FeatureRef(5),
-                    ast.NumVal(6.79699993),
+                    ast.FeatureRef(12),
```
izeigerman, line 361: Not sure I understand why the feature index changed here.

tests/utils.py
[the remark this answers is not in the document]
StrikerRUS: > I'm a little concerned that here we treat a symptom not the cause.

Yeah, you're absolutely right! To fix the root cause we should support multiple floating types in target languages. In this PR I just propose to make our tests a little bit fairer. Native libraries do `double -> float` conversion and in our tests we perform `double -> float -> double`. Unfortunately, I'm not a numerical expert but it seems that `float -> double`  is a safe conversion: https://stackoverflow.com/questions/29648271/convert-float-double-float.

tests/assemblers/test_xgboost.py
StrikerRUS: Oh, all tests have changed due to new train/test splitting routine which is now done by scikit-learn function (`train_test_split`). I found it easier to change `random_state` param only in one place compared to manual `shuffle` in multiple places.

izeigerman: Oh, stupid me. I overlooked that change. It's a bit harder to compare apples to apples because of this though. Do you by chance remember whether there were any changes to expected values before you updated the splitting logic? I'm trying to identify the impact of this update.

StrikerRUS: Sorry, updating splitting logic was the first step as I wanted to play around with `random_state`. Let's choose easier way. Let me split this PR into two.

StrikerRUS: Done in aeb4301dfc88c85edd79e7a8d1304d96a3d35ba9.

izeigerman: Thank you, this is so much better!

PR #281, opened 2020-07-29
run GitHub Actions with Docker

StrikerRUS:
Besides all other benefits, I hoped Docker can help to reduce the runtime of R language test (https://github.com/BayesWitnesses/m2cgen/pull/272#issuecomment-662770082). Unfortunately, it can't,


StrikerRUS:
Possible further steps.

- Upload Docker images to Docker Hub.
- For PRs, download images instead of building them (save ~6min).
- Set trigger to update images at Docker Hub with each commit to `master`.

We use something similar in https://github.com/guolinke/lightgbm-ci-docker.

Issue #283, opened 2020-07-29
add support for class extension and interface in Java generator


yarix:
hi
i would like to have the generated java class to extend or implement an interface. 
the generated class should look something like
```java
package com.acme.ml;
public class MyClassifier implements com.acme.domain.MyInterface {
   public double[] getScore(double[] input) {
   }
}
```
notice the need to remove the `static` modifier from the method getScore

the need is to have multiple models running via calling the interface.

StrikerRUS:
Hello @yarix !

m2cgen tries to generate as minimal as possible working code that can be run from outer piece of code. Refer for example to the following discussion: https://github.com/BayesWitnesses/m2cgen/pull/167#discussion_r379868699.

Maybe I'm missing something but what's the problem to remove `static` word manually after code is generated?

yarix:
Hi @StrikerRUS 
i understand the goal to generate minimal working code, and the ability to manually change the code after it generated.
yet, i think it's useful to let the user generate a class that doesn't need to be manually modified (otherwise, what's the problem to change the class or method name after generation 😉. and yet the tool provide such generators). Of course, this  new functionality has to be optional. 

i made some changes on my fork, will be  glad to contribute back (see https://github.com/BayesWitnesses/m2cgen/compare/master...yarix:master)

i do need some help in removing the `static` in case there is an interface or parent class (I'm not a python developer)

izeigerman:
Hey @yarix thank you so much for reporting this as well as for you PR!

Perhaps I'm missing the point but it seems like I this functionality is somewhat redundant in `m2cgen` and goes beyond what this library is suppose to do.

I'd argue that the manual modification of the generated code is not required either. In OOP world the problem of _adapting_ a one interface to another has been long solved by the [Adapter](https://en.wikipedia.org/wiki/Adapter_pattern) (aka Wrapper) design patter. 

For example in the application code it's fairly easy to do the following:
```
class MyClassifierAdapter implements com.acme.domain.MyInterface {
   public double[] getScore(double[] input) {
      return MyClassifier.score(input)
   }
}
```
This entity has to be created once and can never be revisited even if the model changes.

Alternatively this can be solved in an even more straightforward way if the `MyInterface` is a SAM (Single Abstract Method) interface and you're using Java >=8:
```
MyInterface classifier = (input) -> MyClassifier.score(input)
```
This is even less verbose.

Given the last example, I don't see much value in bringing interface support to `m2cgen`.

Please let us know what you think!

[...]

PR #284, opened 2020-07-29
adding optional parent class and interface to Java Generator 

yarix:
in response to #283


StrikerRUS:
Thank you for your contribution!

Closing according to https://github.com/BayesWitnesses/m2cgen/issues/283#issuecomment-818092505.

PR #288, opened 2020-08-05
simplify linear assemblers and move custom java interpreter tests from bottom for the consistency with other files

StrikerRUS:


tests/interpreters/test_java.py
```diff
@@ -272,129 +272,6 @@ def test_bin_vector_num_expr():
     utils.assert_code_equal(interpreter.interpret(expr), expected_code)
 
 
-def test_abs_expr():
```
izeigerman, line 275: Wow, what happened with this diff? :)

StrikerRUS: I just moved `test_depth_threshold_with_bin_expr`, `test_depth_threshold_without_bin_expr`, `test_deep_mixed_exprs_not_reaching_threshold` and `test_deep_mixed_exprs_exceeding_threshold` tests to the place where they are in all other files. Namely, after `test_bin_vector_num_expr` and before `test_abs_expr`. So now the order of tests are the same across all `test_<language>.py` files. For example, thanks to the same order it allowed to spot some missing tests: #153. No any modifications were done in this file, just simple copy-paste.

PR #290, opened 2020-08-07
Handle missing values replacement in LightGBM

Aulust:
Sometimes exported LGBMRegressor model's prediction doesn't match predictions from the original model. This happens when model encounters values missing during training. More detailed discussion could be found here https://github.com/microsoft/LightGBM/issues/2921 

This is by no means a complete fix for the problem, it only addresses this part of the LightGBM behavior:
"for numerical features, if not missing is seen in training, the missing value will be converted to zero, and then check it with the threshold. So it is not always the left side."

Fix has also being tested on fairly big regression model with numerical features and it works as expected.

How to reproduce:
````
import numpy as np
import lightgbm as lgb
import m2cgen as m2c
from sklearn.datasets import load_diabetes

dataset = load_diabetes()

gbm = lgb.LGBMRegressor(num_leaves=51,
                        learning_rate=0.05,
                        n_estimators=100)
gbm.fit(dataset['data'], dataset['target'])


test = np.array([-2.175, 0.797, np.NaN, 1.193, 0.0, 0.0, 0.0, np.NaN, np.NaN, np.NaN])

print(gbm.predict(np.array([test]))[0])

code = m2c.export_to_python(gbm)

with open('model.py', 'w') as fp:
    fp.write(code)

import model as m

print(m.score(test))
````

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,3 +50,10 @@ RUN update-alternatives --install /usr/bin/python python /usr/bin/python${python
     python -m pip install --upgrade pip && \
     pip install --no-cache-dir Cython numpy && \
     pip install --no-cache-dir -r requirements-test.txt
+
+ENV MKL_NUM_THREADS=2
+ENV NUMEXPR_NUM_THREADS=2
+ENV OMP_NUM_THREADS=2
+ENV OPENBLAS_NUM_THREADS=2
+ENV VECLIB_MAXIMUM_THREADS=2
+ENV BLIS_NUM_THREADS=2
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -317,9 +317,14 @@ class LightGBMModelAssembler(BaseTreeBoostingAssembler):
         op = ast.CompOpType.from_str_op(tree["decision_type"])
         assert op == ast.CompOpType.LTE, "Unexpected comparison op"
 
-        # Make sure that if the "default_left" is true the left tree branch
-        # ends up in the "else" branch of the ast.IfExpr.
-        if tree["default_left"]:
+        missing_type = tree['missing_type']
+
+        if missing_type not in {"NaN", "None"}:
+            raise ValueError(f"Unknown missing_type: {missing_type}")
+
+        reverse_condition = missing_type == "NaN" and tree["default_left"]
+        reverse_condition |= missing_type == "None" and tree["threshold"] >= 0
+        if reverse_condition:
             op = ast.CompOpType.GT
             true_child = tree["right_child"]
             false_child = tree["left_child"]
--- a/tests/e2e/executors/php.py
+++ b/tests/e2e/executors/php.py
@@ -47,7 +47,8 @@ class PhpExecutor(base.BaseExecutor):
         exec_args = [self._php,
                      "-f",
                      file_name,
-                     *map(interpreters.utils.format_float, X)]
+                     "--",
+                     *map(utils.format_arg, X)]
         return utils.predict_from_commandline(exec_args)
 
     def prepare(self):
--- a/tests/e2e/test_e2e.py
+++ b/tests/e2e/test_e2e.py
@@ -34,6 +34,8 @@ HASKELL = pytest.mark.haskell
 RUBY = pytest.mark.ruby
 F_SHARP = pytest.mark.f_sharp
 REGRESSION = pytest.mark.regr
+REGRESSION_WITH_MISSING_VALUES = pytest.mark.regr_missing_val
+CLASSIFICATION_WITH_MISSING_VALUES = pytest.mark.clf_missing_val
 CLASSIFICATION = pytest.mark.clf
 
 
@@ -92,6 +94,32 @@ def regression_bounded(model, test_fraction=0.02):
         model,
         utils.get_bounded_regression_model_trainer(test_fraction),
         REGRESSION,
+    )
+
+
+def regression_w_missing_values(model, test_fraction=0.02):
+    return (
+        model,
+        utils.get_regression_w_missing_values_model_trainer(test_fraction),
+        REGRESSION_WITH_MISSING_VALUES,
+    )
+
+
+def classification_random_w_missing_values(model, test_fraction=0.02):
+    return (
+        model,
+        utils.get_classification_random_w_missing_values_model_trainer(
+            test_fraction),
+        CLASSIFICATION_WITH_MISSING_VALUES,
+    )
+
+
+def classification_binary_random_w_missing_values(model, test_fraction=0.02):
+    return (
+        model,
+        utils.get_classification_binary_random_w_missing_values_model_trainer(
+            test_fraction),
+        CLASSIFICATION_WITH_MISSING_VALUES,
     )
 
 
@@ -185,6 +213,14 @@ STATSMODELS_LINEAR_REGULARIZED_PARAMS = dict(method="elastic_net",
             lightgbm.LGBMClassifier(**LIGHTGBM_PARAMS_LARGE)),
         classification_binary_random(
             lightgbm.LGBMClassifier(**LIGHTGBM_PARAMS_LARGE)),
+
+        # LightGBM (Missing values during train)
+        regression_w_missing_values(
+            lightgbm.LGBMRegressor(**LIGHTGBM_PARAMS)),
+        classification_random_w_missing_values(
+            lightgbm.LGBMClassifier(**LIGHTGBM_PARAMS)),
+        classification_binary_random_w_missing_values(
+            lightgbm.LGBMClassifier(**LIGHTGBM_PARAMS)),
 
         # LightGBM (Different Objectives)
         regression(lightgbm.LGBMRegressor(
@@ -549,6 +585,10 @@ STATSMODELS_LINEAR_REGULARIZED_PARAMS = dict(method="elastic_net",
         classification_binary(
             ensemble.RandomForestClassifier(**FOREST_PARAMS)),
     ],
+    [
+        (R, REGRESSION_WITH_MISSING_VALUES),
+        (R, CLASSIFICATION_WITH_MISSING_VALUES),
+    ]
 
     # Following is the list of extra tests for languages/models which are
     # not fully supported yet.
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -24,6 +24,7 @@ from xgboost import XGBClassifier
 
 from m2cgen import ast
 from m2cgen.assemblers import _get_full_model_name
+from m2cgen.interpreters.utils import format_float
 
 
 class StatsmodelsSklearnLikeWrapper(BaseEstimator, RegressorMixin):
@@ -69,6 +70,7 @@ class ModelTrainer:
     def __init__(self, dataset_name, test_fraction):
         self.dataset_name = dataset_name
         self.test_fraction = test_fraction
+        additional_test_data = None
         np.random.seed(seed=7)
         if dataset_name == "boston":
             self.name = "train_model_regression"
@@ -77,6 +79,12 @@ class ModelTrainer:
             self.name = "train_model_regression_bounded"
             self.X, self.y = datasets.load_boston(return_X_y=True)
             self.y = np.arctan(self.y) / np.pi + 0.5  # (0; 1)
+        elif dataset_name == "diabetes":
+            self.name = "train_model_regression_w_missing_values"
+            self.X, self.y = datasets.load_diabetes(return_X_y=True)
+            additional_test_data = np.array([
+                [np.NaN] * self.X.shape[1],
+            ])
         elif dataset_name == "iris":
             self.name = "train_model_classification"
             self.X, self.y = datasets.load_iris(return_X_y=True)
@@ -93,17 +101,36 @@ class ModelTrainer:
             N = 1000
             self.X = np.random.random(size=(N, 200))
             self.y = np.random.randint(3, size=(N,))
+        elif dataset_name == "classification_rnd_w_missing_values":
+            self.name = "train_model_classification_rnd_w_missing_values"
+            N = 100
+            self.X = np.random.random(size=(N, 20)) - 0.5
+            self.y = np.random.randint(3, size=(N,))
+            additional_test_data = np.array([
+                [np.NaN] * self.X.shape[1],
+            ])
         elif dataset_name == "classification_binary_rnd":
             self.name = "train_model_classification_binary_random_data"
             N = 1000
             self.X = np.random.random(size=(N, 200))
             self.y = np.random.randint(2, size=(N,))
+        elif dataset_name == "classification_binary_rnd_w_missing_values":
+            self.name = \
+                "train_model_classification_binary_rnd_w_missing_values"
+            N = 100
+            self.X = np.random.random(size=(N, 20)) - 0.5
+            self.y = np.random.randint(2, size=(N,))
+            additional_test_data = np.array([
+                [np.NaN] * self.X.shape[1],
+            ])
         else:
             raise ValueError(f"Unknown dataset name: {dataset_name}")
 
         (self.X_train, self.X_test,
-         self.y_train, self.y_test) = train_test_split(
+         self.y_train, _) = train_test_split(
             self.X, self.y, test_size=test_fraction, random_state=13)
+        if additional_test_data is not None:
+            self.X_test = np.vstack((additional_test_data, self.X_test))
 
     @classmethod
     def get_instance(cls, dataset_name, test_fraction=0.02):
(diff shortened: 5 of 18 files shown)

### CODE REVIEW

File: m2cgen/assemblers/boosting.py [hunk 1]
```diff
@@ -317,15 +317,28 @@ def _assemble_tree(self, tree):
         op = ast.CompOpType.from_str_op(tree["decision_type"])
         assert op == ast.CompOpType.LTE, "Unexpected comparison op"
 
-        # Make sure that if the "default_left" is true the left tree branch
-        # ends up in the "else" branch of the ast.IfExpr.
-        if tree["default_left"]:
-            op = ast.CompOpType.GT
-            true_child = tree["right_child"]
-            false_child = tree["left_child"]
+        missing_type = tree['missing_type']
```
izeigerman, line 320: I wonder whether we can eliminate the logic duplication here by merging 2 conditions into one:
```
if missing_type in ("NaN", "None"):
    reverse_condition = missing_type == "NaN" and tree["default_left"]
    reverse_condition |= missing_type == "None" and tree["threshold"] >= 0
    if reverse_condition:
        op = ast.CompOpType.GT
        true_child = tree["right_child"]
        false_child = tree["left_child"]
    else:
        true_child = tree["left_child"]
        false_child = tree["right_child"]
else:
    raise ValueError('Unknown missing_type')
```
Please also note that we use double quotes for string literals.

StrikerRUS:
@Aulust Thank you very much for the fix!

> This is by no means a complete fix for the problem, it only addresses this part of the LightGBM behavior:

Can you please share your thoughts about how hard will it be to implement the full support of `missing_type` logic?

https://github.com/microsoft/LightGBM/blob/33af069cc174a4c56c5f1e2f351196a284248f80/src/io/tree.cpp#L337-L358

m2cgen/assemblers/boosting.py
Aulust: Got it. I'll update pull request with fixes and unit test in a couple of days.

Aulust:
Haven't actually tried to support all use cases, just wanted to fix a problem at hand. I think l'll try to add full support for missing values in a follow up pull request,  but no guarantees. Testing all possible combinations of parameters and values would be challenging

m2cgen/assemblers/boosting.py
StrikerRUS: Thanks! Unit test you can place in the following file:
https://github.com/BayesWitnesses/m2cgen/blob/train_test_split/tests/assemblers/test_lightgbm.py

StrikerRUS:
> but no guarantees

No problem! I can work on this later. I believe it should be enough to introduce some new requirements for target languages: `IsNan` function and logical operators used in `if`s to not overcomplicate already large enough output code from boosting models.

.github/workflows/main.yml
```diff
@@ -34,6 +34,11 @@ jobs:
           TEST: API
           PYTHON: ${{ matrix.python }}
           BUILD_DIRECTORY: ${{ github.workspace }}
+          MKL_NUM_THREADS: 2
```
Aulust, line 37: Without this training results are varying considerably depending on the number of cores.

Setting it to 1 would be more sensible, but for some bizarre reason doing so causes lightning_KernelSVC kernel="poly" test to fail consistently with difference in 4th decimal digit. It seams like a bug, but for now I'll leave it as 2 as this is the number of cores github actions vm has

tests/e2e/executors/php.py
```diff
@@ -47,7 +47,8 @@ def predict(self, X):
         exec_args = [self._php,
                      "-f",
                      file_name,
-                     *map(interpreters.utils.format_float, X)]
+                     "--",
```
Aulust, line 50: php executor fails when first argument is negative

tests/e2e/test_e2e.py
```diff
@@ -549,6 +566,9 @@ def regression_bounded(model, test_fraction=0.02):
         classification_binary(
             ensemble.RandomForestClassifier(**FOREST_PARAMS)),
     ],
+    [
+        (R, REGRESSION_WITH_MISSING_VALUES),
```
Aulust, line 570: Code, generated for R fails on Nan input with
````
Error in if (NaN <= 1) { : missing value where TRUE/FALSE needed
````
skipping for now

File: Dockerfile [hunk 5]
```diff
@@ -50,3 +50,9 @@ RUN update-alternatives --install /usr/bin/python python /usr/bin/python${python
     python -m pip install --upgrade pip && \
     pip install --no-cache-dir Cython numpy && \
     pip install --no-cache-dir -r requirements-test.txt
+
+ENV MKL_NUM_THREADS=2
+ENV NUMEXPR_NUM_THREADS=2
+ENV OMP_NUM_THREADS=2
+ENV OPENBLAS_NUM_THREADS=2
+ENV VECLIB_MAXIMUM_THREADS=2
```
Aulust, line 58: Without this training results are varying considerably depending on the number of cores.

Setting it to 1 would be more sensible, but for some bizarre reason doing so causes lightning_KernelSVC kernel="poly" test to fail consistently with difference in 4th decimal digit. It seams like a bug, but for now I'll leave it as 2 as this is the number of cores github actions vm has

tests/utils.py
```diff
@@ -286,3 +303,10 @@ def inner(*args, **kwarg):
 
 def _is_float(value):
     return isinstance(value, (float, np.floating))
+
+
+def format_float(value):
+    if np.isnan(value):
+        return "NaN"
```
izeigerman, line 310: Just wondering why couldn't we extend the existing function (https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/interpreters/utils.py#L29) with this logic?

tests/e2e/test_e2e.py
izeigerman: @StrikerRUS quick research didn't expose an obvious way to work this around. Neither `NA` nor `NaN` behave in the expected way in comparison expressions. Is there any chance you're aware of some workaround that can help us out here?

izeigerman:
@Aulust this is a fantastic work - thank you so much for introducing a test and variety of other improvements. 

@StrikerRUS I'd appreciate another pair of eyes on this update. Thank you!

tests/utils.py
Aulust: Output of this two functions in theory is different. https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/interpreters/utils.py#L29 is used for model code generation and should be language-specific, for example, returning Float.NaN is case of java. Yet nan seems to never be present in actual model, so no need to modify that function at least for now.

Here format_float generates representation for command-line string parsers, and "NaN" works for all supported languages.

izeigerman: Thanks for clarifying!

Dockerfile
StrikerRUS: > Without this training results are varying considerably depending on the number of cores.

Could you please recollect in memory which libraries were affected by the different number of threads? 

Also, please add `BLIS_NUM_THREADS` env variable here.

m2cgen/assemblers/boosting.py
```diff
@@ -317,9 +317,14 @@ def _assemble_tree(self, tree):
         op = ast.CompOpType.from_str_op(tree["decision_type"])
         assert op == ast.CompOpType.LTE, "Unexpected comparison op"
 
-        # Make sure that if the "default_left" is true the left tree branch
-        # ends up in the "else" branch of the ast.IfExpr.
-        if tree["default_left"]:
+        missing_type = tree['missing_type']
+
+        if missing_type not in ("NaN", "None"):
```
StrikerRUS, line 322:
```suggestion
        if missing_type not in {"NaN", "None}:
```

m2cgen/assemblers/boosting.py
```diff
@@ -317,9 +317,14 @@ def _assemble_tree(self, tree):
...
+            raise ValueError('Unknown missing_type')
```
StrikerRUS, line 323:
```suggestion
            raise ValueError(f"Unknown missing_type: {missing_type}")
```

tests/e2e/test_e2e.py
```diff
@@ -158,23 +167,31 @@ def regression_bounded(model, test_fraction=0.02):
...
+        regression_with_missing_values(
+            lightgbm.LGBMRegressor(**LIGHTGBM_PARAMS_GOSS)),
         classification(lightgbm.LGBMClassifier(**LIGHTGBM_PARAMS_GOSS)),
         classification_binary(lightgbm.LGBMClassifier(
             **LIGHTGBM_PARAMS_GOSS)),
 
         # LightGBM (RF)
         regression(lightgbm.LGBMRegressor(**LIGHTGBM_PARAMS_RF)),
+        regression_with_missing_values(
+            lightgbm.LGBMRegressor(**LIGHTGBM_PARAMS_RF)),
```
StrikerRUS, line 194: Do we really need to test each LightGBM's boosting type (DART, GOSS, RF) with missing values? If I'm not mistaken, missing values are handled in the same way by the same code in all boosting types. So, I believe it will be enough to test only one algorithm. 

I think it will be better to add classification case instead.

Also, please add a separate group for regression with missing values to not break the logical grouping of `regression, classification, classification binary` which is used across the whole file for all supported libraries.

tests/utils.py
```diff
@@ -77,6 +78,12 @@ def __init__(self, dataset_name, test_fraction):
             self.name = "train_model_regression_bounded"
             self.X, self.y = datasets.load_boston(True)
             self.y = np.arctan(self.y) / np.pi + 0.5  # (0; 1)
+        elif dataset_name == "diabetes":
+            self.name = "train_model_regression_with_missing_values"
+            self.X, self.y = datasets.load_diabetes(True)
```
StrikerRUS, line 83: Refer to #294.
```suggestion
            self.X, self.y = datasets.load_diabetes(return_X_y=True)
```

StrikerRUS (changes requested):
@Aulust Thank you very much for the fix! Please consider addressing some minor comments below. Also, I left some comments inline in already started threads to not split discussions.

Dockerfile
StrikerRUS: > but for some bizarre reason doing so causes lightning_KernelSVC kernel="poly" test to fail consistently with difference in 4th decimal digit.

I see the same thing are happening in #271. Probably it is related to numpy's random generator - they have just changed it in `1.19`.

tests/utils.py
StrikerRUS: > Yet nan seems to never be present in actual model, so no need to modify that function at least for now.

I agree with this statement! But also we should avoid code duplication. It is possible we will need to modify `np.format_float_positional` arguments in the near future quite often, because it was added to the codebase several days ago. With the current design we have to do it in two places.

Please try to reuse the code of `format_float` from `interpreters/utils.py` with decorator or simply 
```
if np.isnan(value):
    return "NaN"
else:
    return m2cgen.interpreters.utils.format_float(value)
```

Also, I guess it will be better to have a new unique name for this function. Something like `format_float_with_nan`.

Dockerfile
Aulust: On my machine only test_lightgbm.py::test_regression_random_forest fails without this environment variables with
`AssertionError: 20.182372860039543 != 36.40634951405711`

tests/e2e/test_e2e.py
Aulust: I've added tests for each boosting type just for the sake of grouping, it's most likely not necessary. Agree, it would be better to make a separate group for such tests.

tests/e2e/test_e2e.py
StrikerRUS: Indeed, in R you got `NA` when compare `NaN`s
```
print(NaN <= 1)  # [1] NA
```

It is possible to wrap condition into `isTRUE()` function, but I'm not sure whether it is the right way:
```
print(isTRUE(NaN <= 1))  # [1] FALSE
```

m2cgen/assemblers/boosting.py
izeigerman: Missing double-quote after `None`

Aulust:
@StrikerRUS , please take a look at updated pull request with fixes

tests/utils.py
```diff
@@ -93,17 +101,35 @@ def __init__(self, dataset_name, test_fraction):
             N = 1000
             self.X = np.random.random(size=(N, 200))
             self.y = np.random.randint(3, size=(N,))
+        elif dataset_name == "classification_rnd_w_missing_values":
+            self.name = "train_model_classification_rnd_w_missing_values"
+            N = 1000
+            self.X = np.random.random(size=(N, 200)) - 0.5
```
StrikerRUS, line 107: I believe `1000x200` is too much for this test. `1000x200` was used in random data to test very large trees: #105.

tests/utils.py
```diff
@@ -93,17 +101,35 @@ def __init__(self, dataset_name, test_fraction):
...
+            additional_test_data = np.array([
+                [np.NaN] * self.X.shape[1],
+            ])
         elif dataset_name == "classification_binary_rnd":
             self.name = "train_model_classification_binary_random_data"
             N = 1000
             self.X = np.random.random(size=(N, 200))
             self.y = np.random.randint(2, size=(N,))
+        elif dataset_name == "classification_binary_rnd_w_missing_values":
+            self.name = "train_model_classification_binary_rnd_w_missing"
```
StrikerRUS, line 118:
```suggestion
            self.name = "train_model_classification_binary_rnd_w_missing_values"
```

tests/utils.py
```diff
@@ -93,17 +101,35 @@ def __init__(self, dataset_name, test_fraction):
...
+            N = 1000
+            self.X = np.random.random(size=(N, 200)) - 0.5
```
StrikerRUS, line 120: The same as above about number of data points and features.

tests/utils.py
```diff
@@ -201,10 +227,23 @@ def assert_code_equal(actual, expected):
...
+
 get_bounded_regression_model_trainer = functools.partial(
     ModelTrainer.get_instance, "boston_y_bounded")
 
 
+get_regression_with_missing_values_model_trainer = functools.partial(
+    ModelTrainer.get_instance, "diabetes")
+
+
+get_classification_random_w_missing_values_trainer = functools.partial(
```
StrikerRUS, line 239:
```suggestion
get_classification_random_w_missing_values_model_trainer = functools.partial(
```

tests/utils.py
```diff
@@ -201,10 +227,23 @@ def assert_code_equal(actual, expected):
...
+    ModelTrainer.get_instance, "classification_rnd_w_missing_values")
+
+
+get_classification_binary_random_w_missing_values_trainer = functools.partial(
```
StrikerRUS, line 243:
```suggestion
get_classification_binary_random_w_missing_values_model_trainer = functools.partial(
```

StrikerRUS (commented):
Thanks @Aulust !

Looks much better now!

Just some minor comments about the consistency in names and about the data size.


I'm just worried about the R lang. Have you tried to find any better solution than https://github.com/BayesWitnesses/m2cgen/pull/290#discussion_r476062643?

tests/utils.py
Aulust: I tried to fit names in a single line, hence the inconsistency in naming. I'll try your suggestions

Aulust:
As for R, I believe a better solution would be what you mentioned considering handling all possible types of missing values:
> I believe it should be enough to introduce some new requirements for target languages: IsNan function and logical operators used in ifs

Do you want a fix for R as a part of this pull request?

tests/utils.py
StrikerRUS: Ah, I forgot that we should follow that outmoded lint rule of 80 chars per line, sorry! Nevermind!

StrikerRUS:
> Do you want a fix for R as a part of this pull request?

Not actually, just wanted to know your opinion and your finding about R. I think R should not be a blocker for this PR. Especially, as we have a plan to "fix" it with the support of all possible types of missing values. Thank you!

tests/utils.py
Aulust: I've already changed names using \ to break long lines. Would it be better to leave it like this or change back to shortened names?

StrikerRUS: Oh, sorry, I didn't notice that! I'm OK with any solution.

tests/utils.py
```diff
@@ -201,10 +228,25 @@ def assert_code_equal(actual, expected):
 get_classification_binary_random_data_model_trainer = functools.partial(
     ModelTrainer.get_instance, "classification_binary_rnd")
 
+
 get_bounded_regression_model_trainer = functools.partial(
     ModelTrainer.get_instance, "boston_y_bounded")
 
 
+get_regression_with_missing_values_model_trainer = functools.partial(
```
StrikerRUS, line 236: For the consistency with all other names from this PR.

```suggestion
get_regression_w_missing_values_model_trainer = functools.partial(
```

StrikerRUS (approved):
Thanks again for all your hard work and valuable contribution @Aulust !
LGTM!

Aulust:
Thank you

izeigerman (approved):
Thanks for all the hard work @Aulust!

PR #291, opened 2020-08-07
simplify train/test splitting routine for tests

StrikerRUS:


tests/utils.py
```diff
@@ -72,30 +72,22 @@ def __init__(self, dataset_name, test_fraction):
...
-            dataset = datasets.load_breast_cancer()
-            self.X, self.y = shuffle(
-                dataset.data, dataset.target, random_state=13)
+            self.X, self.y = datasets.load_breast_cancer(True)
         elif dataset_name == "regression_rnd":
             self.name = "train_model_regression_random_data"
             N = 1000
             self.X = np.random.random(size=(N, 200))
-            self.y = np.random.random(size=(N, 1))
+            self.y = np.random.random(size=(N,))
```
StrikerRUS, line 90:
```
DataConversionWarning: A column-vector y was passed when a 1d array was expected. Please change the shape of y to (n_samples, ), for example using ravel().
```

PR #295, opened 2020-08-13
update info in the setup.py file

StrikerRUS:


setup.py
```diff
@@ -28,9 +29,10 @@
         "Programming Language :: Python :: 3.6",
         "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
-        "Programming Language :: Python :: Implementation :: CPython",
```
StrikerRUS, line 31: I believe our code can be used with any Python implementation.

setup.py
```diff
@@ -28,9 +29,10 @@
...
     ],
-    keywords="sklearn ml code-generation",
+    keywords=("sklearn statsmodels lightning xgboost lightgbm "
+              "machine-learning ml regression classification "
+              "transpilation code-generation"),
```
StrikerRUS, line 35: Taken from GitHub repo tags:

image

Issue #228, opened 2020-06-01
Add support for probit and cauchy link functions in statsmodels GLM


StrikerRUS:
Refer to https://github.com/BayesWitnesses/m2cgen/pull/195#issuecomment-624350997.

PR #304, opened 2020-08-30
added cauchy link function

StrikerRUS:
Refer to #228.


m2cgen/assemblers/fallback_expressions.py
```diff
@@ -66,6 +64,118 @@ def log1p(expr):
         utils.div(utils.mul(expr, ast.LogExpr(expr1p)), expr1pm1))
 
 
+def atan(expr):
```
izeigerman, line 67: This looks super scary :D

izeigerman (approved):
This looks great, thanks!

m2cgen/assemblers/fallback_expressions.py
StrikerRUS: Yeah, I know! 😃 

But it was the only one I could find with appropriate accuracy. Other either have error about 1e-4 - 1e-6 or look even more scary. I even cannot remember how many trials I've made. There were Chebyshev, CORDIC, Taylor, Remez (in [Sollya package](http://sollya.gforge.inria.fr/sollya-2.9/help.php?name=remez))...

izeigerman: Wow, impressive. Amazing work as always 👍

StrikerRUS: Thank you!

Issue #298, opened 2020-08-21
Error when compiling exported Java code using javac: Too many constants


kijz:
Hi!

I have problems when compiling the generated Java Code from m2cgen.
I am using an XGBClassifier with n_estimators = 400.
The generated code is ~360k lines long, which is around 17MB.

Unfortunately, this code does not compile with javac as i get this error:
```
Model.java:1: error: too many constants
public class Model {
       ^
1 error
```
Using the approach suggested in #297, i have managed to split the score function into subroutines, which unfortunately does not fix my problem.

Is there any workaround for this type of error?
Manually adjusting the file is not really an option since i am planning to create a large number of models and compare them

izeigerman:
Hey @fkajzer! Sorry you run into this issue.

Unlike in the #297 issue, splitting code into subroutines doesn't help in your case, because the `too many constants` error is a constrain of the compiled Java class and not just a single method. 

Unfortunately `m2cgen` doesn't support splitting code into multiple classes / modules today.

The only workaround that I can suggest for now is to reduce the number or size of your estimators. Though I realize that this likely doesn't work for you.

Aulust:
Hi, I've encountered the same problem, suggested fix is in https://github.com/BayesWitnesses/m2cgen/pull/306

Issue #297, opened 2020-08-17
Java Score function is too long. The Java Compiler doesn't allow it to execute. 


sharmahemlata:
Hi, 
I exported an ExtraTreesRegressor Model to java. 
This is the java error I get:
                    The code of method score(double[]) is exceeding the 65535 bytes limit
To get rid of this error, I have to split the function into smaller bits, however the function is thousands of lines long with complex nested if-else loops. It is very time consuming to split the function manually. 

Is there any way the exported code can be refactored or restricted in length? Or  is there any way to work around this error in java.

izeigerman:
Hey @sharmahemlata! Sorry to hear that you run into this issue.

We actually already have a mechanism in place which automatically splits the code into sub-methods based on the size of the AST. Unfortunately the mechanism itself is a heuristics which apparently doesn't work well in all scenario. 
For Java code generator this logic is controlled by the following 2 constants - https://github.com/BayesWitnesses/m2cgen/blob/master/m2cgen/interpreters/java/interpreter.py#L16-L17. 

Unfortunately today we don't expose these parameters in API (although we should consider doing so). As a workaround I'd suggest you to check out the code repo and try reducing these 2 parameters locally one at a time (starting with the `ast_size_per_subroutine_threshold`) until your generated code works. Please let us know how it goes!

izeigerman:
To summarize I suggest the following:

- Try reducing the `ast_size_per_subroutine_threshold ` value to eg. 3000 while keeping the `ast_size_check_frequency` value unchanged.
- If the previous change doesn't work, try reducing the `ast_size_check_frequency` to 50 while keeping the `ast_size_per_subroutine_threshold` value unchanged.
- If the previous approach doesn't work, try reducing both parameters at the same time.

StrikerRUS:
> As a workaround I'd suggest you to check out the code repo and try reducing these 2 parameters

Or it is possible to do it automatically via grid search in your code with help of OOP API:
```
from sklearn.datasets import load_boston
from sklearn.ensemble import ExtraTreesRegressor

import m2cgen as m2c

X, y = load_boston(return_X_y=True)
est = ExtraTreesRegressor(10).fit(X, y)

assemble = m2c.assemblers.RandomForestModelAssembler(est).assemble()

class CustomJavaInterpreter(m2c.interpreters.JavaInterpreter):
    pass

interpreter = CustomJavaInterpreter()
for ast_size_threshold in range(4600, 3000, -500):
[...]

PR #306, opened 2020-09-04
Fix java export for huge models

Aulust:
This is a suggestion on how to tackle a problem mentioned at https://github.com/BayesWitnesses/m2cgen/issues/298 . In my case, I had about 1000 estimators and this fix worked quite well in production.

This fix has some flaws since it brakes backward compatibility in a sense that you would get several class files after compilation. This might be relevant in cases like mine when a custom class loader is used to load new models in runtime without an application restart. A class loader written with only one model class file in mind might not work correctly after this fix.

Another problem comes from R code generator. With huge trees, context stack overflow emerges from nested function calls and I don't know a good fix for that other than refactoring ast tree generation to reduce 

And lastly, I would like to discuss java code performance. Currently, it's quite bad since generated methods are too big to be jit-compiled by default. A simple jmh benchmark as well as checking this in production shows up to 10 times boost in performance with VM flag -XX:-DontCompileHugeMethods. Reducing ast_size_per_subroutine_threshold helps, but it creates more function calls which is expensive (default jvm limit for method inline, FreqInlineSize, is too small to overcome that). After a decent amount of tweaking ast_size_per_subroutine_threshold I was able to get about 80% of -DontCompileHugeMethods performance, but it certainly depends on a particular model.
I see several ways to mitigate this:
- Mentioning in the FAQ about DontCompileHugeMethods or HugeMethodLimit.
- Making ast_size_per_subroutine_threshold a configurable parameter for export_to_java function
- Reducing default ast_size_per_subroutine_threshold to gain some performance (this might also help https://github.com/BayesWitnesses/m2cgen/issues/297)

Would like to hear your take on it.


StrikerRUS:
@Aulust 
I'm so so sorry for the huge delay and for that this project was a kind of abandoned!

I've updated all project dependencies and fixed CI recently. And now I have enough rights to review and merge PRs in this repo.
Are you still interested in finishing this PR?

P.S. Unfortunately, I'm not familiar with Java, but I'll do my best to help with this PR.

Aulust:
Happy to hear that this project is back to being maintained. I've actually been using it in production this whole time (with this fix applied), so the pull request is still valid and I would like it to be ultimately merged into the master branch.
I'll rebase it in a couple of days and check if fixes/updates are required.

StrikerRUS:
> I've actually been using it in production this whole time

This is so cool to hear! 🎉 

> I'll rebase it ...

Many thanks! I'll get familiar with the content of this PR in a few days.

StrikerRUS:
Thanks a lot for this PR! Let me share some my initial thoughts.

I have only one conceptual question about the implementation. Do we want to enhance the current ``SubroutinesMixin`` class or maybe we can use inheritance and introduce new ``ModulesWithSubroutinesMixin`` child class? New class will allow to overcome the problem of the lack of modules in R. Also, for other language generators it may be enough to just split code into subroutines without thinking about modules. Something like "basic" and "advanced" versions of mixin.

> This fix has some flaws since it brakes backward compatibility in a sense that you would get several class files after compilation.

I believe this is totally fine given the benefits that new API introduces.

> Another problem comes from R code generator. With huge trees, context stack overflow emerges from nested function calls ...

Is it something obsolete from the initial implementation? Looks like that new tests with huge trees are passed in R.

I know one awesome R expert and I think I can talk with him about R support in m2cgen. From my point of view, R is the most challenging language to support here :-(.


> ... other than refactoring ast tree generation to reduce recursion.

Do you know any variants? I've spent quite large amount of time reading articles about decision trees inference, but unfortunately haven't found anything useful.

> And lastly, I would like to discuss java code performance.
> Mentioning in the FAQ about DontCompileHugeMethods or HugeMethodLimit.

Awesome research and great idea documenting this in the FAQ! Thank you very much! I fully support it and PR is very welcome. Also, linking #152 here as еру related issue.

> Making ast_size_per_subroutine_threshold a configurable parameter for export_to_java function

To be honest, I don't want overcomplicate `export_to_xxx()` API. I'd like to keep it as simple as possible. Instead, I think we can go more general way and better document OOP API (see #256). So, the majority of users will be happy with very simple `export_to_xxx()` API while those users who need advanced options will use well-documented OOP API. Please share WDYT.

> Reducing default ast_size_per_subroutine_threshold to gain some performance ...

I'm not against it. Feel free to propose a PR!

m2cgen/interpreters/mixins.py
```diff
@@ -112,6 +115,7 @@ class SubroutinesMixin(BaseToCodeInterpreter):
     # disabled by default
     ast_size_check_frequency = sys.maxsize
     ast_size_per_subroutine_threshold = sys.maxsize
+    subroutine_per_group_threshold = sys.maxsize
```
izeigerman, line 118: I feel like `subroutine_per_module_threshold` seem more accurate

m2cgen/interpreters/mixins.py
```diff
@@ -124,15 +128,33 @@ def process_subroutine_queue(self, top_code_generator):
...
         """
         self._subroutine_idx = 0
+        subroutines = []
 
-        while len(self.subroutine_expr_queue):
+        while self.subroutine_expr_queue:
             self._reset_reused_expr_cache()
             subroutine = self.subroutine_expr_queue.pop(0)
             subroutine_code = self._process_subroutine(subroutine)
+            subroutines.append((subroutine, subroutine_code))
```
izeigerman, line 137: If my assumption in the previous comment is correct I don't think we actually need `subroutine` here and having only `subroutine_code` will suffice.

izeigerman (changes requested):
@Aulust , I think this update is absolutely amazing and I really appreciate you revisiting this almost 2 years later! Left a few minor comments but looks fantastic otherwise!

m2cgen/interpreters/mixins.py
```diff
@@ -145,7 +167,16 @@ def _pre_interpret_hook(self, expr, ast_size_check_counter=0, **kwargs):
                 ast_size = ast.count_exprs(expr)
                 if ast_size > self.ast_size_per_subroutine_threshold:
                     function_name = self._get_subroutine_name()
-                    self.enqueue_subroutine(function_name, expr)
+
+                    self.enqueue_subroutine(function_name, self._subroutine_idx, expr)
+
+                    group_idx = self._subroutine_idx // self.subroutine_per_group_threshold
```
izeigerman, line 173: `module_idx`?

izeigerman:
Hey @Aulust! I think this change is great and will be an amazing addition to the library. Any chance you plan to finish the work on this PR?

PR #358, opened 2021-03-31
fix coveralls

StrikerRUS:


StrikerRUS:
Fixed at coveralls.io side by adjusting settings.
Also, disabled annoying bot comments. Now only `coverage/coveralls` check is active.

PR #371, opened 2021-04-08
Support Python 3.9

StrikerRUS:
Depends on https://github.com/scikit-learn-contrib/lightning/pull/154.


StrikerRUS:
> Depends on scikit-learn-contrib/lightning#154.

Merged workaround in #392.

PR #372, opened 2021-04-08
Bump Ubuntu version in Docker

StrikerRUS:
Also, sort installing packages alphabetically.


.github/workflows/main.yml
```diff
@@ -40,6 +40,7 @@ jobs:
             export TEST="E2E";
           fi
           docker run \
+            --ulimit stack=-1 \
```
StrikerRUS, line 43: This is to workaround the following error in R test which didn't occur on Bionic.
```
        stdout, stderr = result.communicate()
        if result.returncode != 0:
            raise Exception(
>               f"Bad exit code ({result.returncode}), "
                f"stderr:\n{stderr.decode('utf-8')}")
E           Exception: Bad exit code (1), stderr:
E           Error: C stack usage  7974932 is too close to the limit
E           Execution halted
...
test_e2e[sklearn_SVR - r_lang - train_model_regression]
```

Dockerfile
```diff
@@ -30,36 +39,27 @@ RUN apt-get update && \
...
+        zulu-8 && \
     rm -rf /var/lib/apt/lists/*
 
 WORKDIR /m2cgen
 
 COPY requirements-test.txt ./
 RUN update-alternatives --install /usr/bin/python python /usr/bin/python${python} 1 && \
     python -m pip install --upgrade pip && \
-    pip install --no-cache-dir Cython numpy && \
+    pip install --no-cache-dir Cython "numpy==1.19.2" && \
```
StrikerRUS, line 64: `lightning` package need to be compiled with the same `numpy` version as one used in tests to prevent the following error:
```
__init__.pxd:242: in init lightning.impl.randomkit.random_fast
    ???
E   ValueError: numpy.ndarray size changed, may indicate binary incompatibility. Expected 88 from C header, got 80 from PyObject
```

StrikerRUS:
I run 5 consequential CI runs to spot any possible flakiness. All of them were green.

PR #387, opened 2021-04-21
Fix XGBoost tests with best_ntree_limit

StrikerRUS:


StrikerRUS:
Integrated into #386.

PR #444, opened 2021-08-07
Add tests for interpreting unsupported exprs

StrikerRUS:


m2cgen/interpreters/interpreter.py
```diff
@@ -55,10 +55,10 @@ def _reset_reused_expr_cache(self):
...
     def _select_handler(self, expr):
         handler_name = _get_handler_name(type(expr))
-        if hasattr(self, handler_name):
-            return getattr(self, handler_name)
-        raise NotImplementedError(
-            f"No handler found for '{type(expr).__name__}'")
+        handler = getattr(self, handler_name, None)
+        if handler is None:
+            raise NotImplementedError(f"No handler found for '{type(expr).__name__}'")
+        return handler
```
StrikerRUS, line 61: Speedup code generation by removing unnecessary `hasattr()` check.
> Checking if an attribute is present (using `hasattr`) is in general just as expensive as getting the attribute (`getattr` or dot notation).
   https://scikit-learn.org/stable/glossary.html?highlight=getattr#term-duck-typing

PR #511, opened 2022-03-27
Significantly reduce the number of redundant parentheses in the generated code

izeigerman:
Admittedly the original idea of wrapping each operand of an infix expression into parentheses was a half-assed solution. 
I've noticed recurrent complains from users who mentioned bloated file size, crashed editors, etc. So I figured I try and introduce the proper operator precedence to the code generation logic and only add parentheses where they are necessary.

### DIFF
--- a/m2cgen/ast.py
+++ b/m2cgen/ast.py
@@ -7,6 +7,7 @@ import numpy as np
 
 class Expr:
     output_size = 1
+    precedence = None
     # Setting this value to true serves as an indication that the result
     # of evaluation of this expression is being used in other expressions
     # and it's recommended to persist or cache it in some way.
@@ -206,6 +207,8 @@ class TanhExpr(NumExpr):
 
 
 class PowExpr(NumExpr):
+    precedence = 4
+
     def __init__(self, base_expr, exp_expr, to_reuse=False):
         assert base_expr.output_size == 1, "Only scalars are supported"
         assert exp_expr.output_size == 1, "Only scalars are supported"
@@ -243,6 +246,7 @@ class BinNumExpr(NumExpr, BinExpr):
         self.right = right
         self.op = op
         self.to_reuse = to_reuse
+        self.precedence = 3 if op in (BinNumOpType.MUL, BinNumOpType.DIV) else 2
 
     def __str__(self):
         return (f"BinNumExpr({self.left},{self.right},{self.op.name},"
@@ -372,6 +376,8 @@ COMP_OP_TYPE_MAPPING = {e.value: e for e in CompOpType}
 
 
 class CompExpr(BoolExpr):
+    precedence = 1
+
     def __init__(self, left, right, op):
         assert left.output_size == 1, "Only scalars are supported"
         assert right.output_size == 1, "Only scalars are supported"
--- a/m2cgen/interpreters/code_generator.py
+++ b/m2cgen/interpreters/code_generator.py
@@ -99,10 +99,11 @@ class BaseCodeGenerator:
             lines = lines.strip().split(new_line)
         self._write_to_code_buffer(f"{new_line.join(lines)}{new_line}", prepend=True)
 
-    # Following methods simply compute expressions using templates without changing result.
+    def infix_expression(self, left, right, op, wrap=False):
+        result = self.tpl_infix_expression(left=left, right=right, op=op)
+        return result if not wrap else f"({result})"
 
-    def infix_expression(self, left, right, op):
-        return self.tpl_infix_expression(left=left, right=right, op=op)
+    # Following methods simply compute expressions using templates without changing result.
 
     def num_value(self, value):
         return self.tpl_num_value(value=value)
@@ -188,7 +189,7 @@ class CLikeCodeGenerator(ImperativeCodeGenerator):
     """
 
     tpl_num_value = CodeTemplate("{value}")
-    tpl_infix_expression = CodeTemplate("({left}) {op} ({right})")
+    tpl_infix_expression = CodeTemplate("{left} {op} {right}")
     tpl_var_declaration = CodeTemplate("{var_type} {var_name};")
     tpl_return_statement = CodeTemplate("return {value};")
     tpl_array_index_access = CodeTemplate("{array_name}[{index}]")
--- a/m2cgen/interpreters/interpreter.py
+++ b/m2cgen/interpreters/interpreter.py
@@ -1,9 +1,12 @@
 from m2cgen.assemblers import fallback_expressions
-from m2cgen.ast import IfExpr
+from m2cgen.ast import IfExpr, BinNumExpr, CompExpr
 from m2cgen.interpreters.utils import CachedResult, _get_handler_name
 
 
 class BaseInterpreter:
+
+    infix_expressions = [BinNumExpr, CompExpr]
+
     """
     Base class of AST interpreter. Provides single public method .interpret()
     which takes instance of AST expression and recursively applies method
@@ -21,7 +24,7 @@ class BaseInterpreter:
     def _pre_interpret_hook(self, expr, **kwargs):
         return None, kwargs
 
-    def _do_interpret(self, expr, to_reuse=None, **kwargs):
+    def _do_interpret(self, expr, to_reuse=None, left_precedence=None, right_precedence=None, **kwargs):
         # Hook which allows to override kwargs and to return custom result.
         result, kwargs = self._pre_interpret_hook(expr, **kwargs)
 
@@ -34,6 +37,15 @@ class BaseInterpreter:
 
         handler = self._select_handler(expr)
 
+        # Reset precedence on non-infix expressions.
+        if not isinstance(expr, tuple(self.infix_expressions)):
+            left_precedence = None
+            right_precedence = None
+
+        result = handler(
+            expr, left_precedence=left_precedence,
+            right_precedence=right_precedence, **kwargs)
+
         # Note that the reuse flag passed in the arguments has a higher
         # precedence than one specified in the expression. One use case for
         # this behavior is to override the original to_reuse flag for
@@ -41,9 +53,8 @@ class BaseInterpreter:
         # subroutines are not supported by specific interpreter implementation.
         expr_to_reuse = to_reuse if to_reuse is not None else expr.to_reuse
         if not expr_to_reuse:
-            return handler(expr, **kwargs)
+            return result
 
-        result = handler(expr, **kwargs)
         return self._cache_reused_expr(expr, result)
 
     def _cache_reused_expr(self, expr, expr_result):
@@ -107,11 +118,12 @@ class ToCodeInterpreter(BaseToCodeInterpreter):
             op=op,
             right=self._do_interpret(expr.right, **kwargs))
 
-    def interpret_bin_num_expr(self, expr, **kwargs):
+    def interpret_bin_num_expr(self, expr, left_precedence=None, right_precedence=None, **kwargs):
         return self._cg.infix_expression(
-            left=self._do_interpret(expr.left, **kwargs),
+            left=self._do_interpret(expr.left, left_precedence=expr.precedence, **kwargs),
             op=expr.op.value,
-            right=self._do_interpret(expr.right, **kwargs))
+            right=self._do_interpret(expr.right, right_precedence=expr.precedence, **kwargs),
+            wrap=self._wrap_infix_expr(expr, left_precedence, right_precedence))
 
     def interpret_num_val(self, expr, **kwargs):
         return self._cg.num_value(value=expr.value)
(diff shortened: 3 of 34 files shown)


izeigerman:
Well, actually, the infix power operator seems to be common enough to deserve a mixin of its own or some other means of generalization. Will refactor this.

izeigerman:
@StrikerRUS I think this should be ready for review now. Thanks!

izeigerman:
I'm going to regenerate code examples in a follow-up PR.

StrikerRUS:
I'm so excited you've come back to this project!

Also I would like to sorry for not allocating significant amount of time and energy to this project recently.

m2cgen/interpreters/interpreter.py
```diff
@@ -190,13 +201,10 @@ def interpret_tanh_expr(self, expr, **kwargs):
...
-    def interpret_pow_expr(self, expr, **kwargs):
-        if self.power_function_name is NotImplemented:
-            raise NotImplementedError("Power function is not provided")
-        self.with_math_module = True
-        base_result = self._do_interpret(expr.base_expr, **kwargs)
-        exp_result = self._do_interpret(expr.exp_expr, **kwargs)
-        return self._cg.function_invocation(self.power_function_name, base_result, exp_result)
+    def _wrap_infix_expr(self, expr, left_precedence, right_precedence):
+        wrap = left_precedence is not None and left_precedence > expr.precedence
+        wrap |= right_precedence is not None and right_precedence >= expr.precedence
```
StrikerRUS, line 206: Maybe we can avoid `is not None` by changing default precedence (see comment above)?

m2cgen/interpreters/mixins.py
```diff
@@ -195,3 +195,46 @@ def _get_subroutine_name(self):
...
+    This mixin is used for languages that provide the exponentiation operation in a form
+    of an infix operator. Such languages inlcude Haskell, F#, R and others.
+
+    The operator used by default is "**", but it can be overriden by setting an appropriate
+    value in `pow_operator`.
+    """
+
+    pow_operator = "**"
+
+    infix_expressions = [*BaseToCodeInterpreter.infix_expressions, ast.PowExpr]
```
StrikerRUS, line 211: See comment above about eliminating `list -> tuple` cast for some speedup.
```suggestion
    infix_expressions = (*BaseToCodeInterpreter.infix_expressions, ast.PowExpr)
```

m2cgen/interpreters/mixins.py
```diff
@@ -195,3 +195,46 @@ def _get_subroutine_name(self):
...
+
+    def interpret_pow_expr(self, expr, left_precedence=None,
+                           right_precedence=None, **kwargs):
+        base_result = self._do_interpret(
+            expr.base_expr, left_precedence=expr.precedence, **kwargs)
+        exp_result = self._do_interpret(
+            expr.exp_expr, right_precedence=expr.precedence, **kwargs)
+        return self._cg.infix_expression(
+            left=base_result, right=exp_result, op=self.pow_operator,
+            wrap=self._wrap_infix_expr(expr, left_precedence, right_precedence))
```
StrikerRUS, line 221: Same as above about code readability.
```suggestion
        return self._cg.infix_expression(
            left=base_result,
            right=exp_result,
            op=self.pow_operator,
            wrap=self._wrap_infix_expr(expr, left_precedence, right_precedence))
```

m2cgen/interpreters/mixins.py
```diff
@@ -195,3 +195,46 @@ def _get_subroutine_name(self):
 
     def create_code_generator(self):
         raise NotImplementedError
+
+
+class InfixPowExprMixin(BaseToCodeInterpreter):
```
StrikerRUS, line 200: IMHO, `PowExprInfixMixin` and `PowExprFunctionMixin` names are slightly better because the first word highlights a context about what these classes are (they are about for to implement power expression).

Also, it would be easier for contributors to search for appropriate mixin for a new programming language.

Just as an example: https://docs.opencv.org/3.4/d8/d01/group__imgproc__color__conversions.html

Note enumerations are in a form of `COLOR_xxx` but not in a form of `xxx_COLOR`.

m2cgen/interpreters/mixins.py
```diff
@@ -195,3 +195,46 @@ def _get_subroutine_name(self):
...
+        base_result = self._do_interpret(
+            expr.base_expr, left_precedence=expr.precedence, **kwargs)
+        exp_result = self._do_interpret(
+            expr.exp_expr, right_precedence=expr.precedence, **kwargs)
+        return self._cg.infix_expression(
+            left=base_result, right=exp_result, op=self.pow_operator,
+            wrap=self._wrap_infix_expr(expr, left_precedence, right_precedence))
+
+
+class FunctionPowExprMixin(BaseToCodeInterpreter):
```
StrikerRUS, line 224: Maybe make one of these two mixins as a default behavior? This will allow contributors to not get familiar with mixins mechanism in a half (roughly, of course) of all cases.

StrikerRUS (changes requested):
Just awesome simplification! Really love it!
Please consider checking some my minor comments below.

izeigerman:
> Also I would like to sorry for not allocating significant amount of time and energy to this project recently.

@StrikerRUS you have absolutely nothing to be sorry about! I really appreciate all the time and effort you've put into this project. Thank you and thank you for the great review as always 👍  Working through your comments now

m2cgen/interpreters/mixins.py
izeigerman: I actually didn't know this syntax was supported! Otherwise I wouldn't have introduced the list->tuple conversion. TIL!

m2cgen/interpreters/mixins.py
izeigerman: Yeah, I feel like we need to keep up with modern best practices and introduce auto-formatting to the project (eg. `black`). Will address this manually for now and will introduce `black` support in future PRs.

m2cgen/interpreters/mixins.py
izeigerman: So I actually did this intentionally :) I want the developer to make a conscious choice and pick one or the other. 

> Maybe make one of these two mixins as a default behavior? 

Well, I don't think there is such thing as a "default" behavior in this case, since even with the pow expression as a function a potential contributor needs to provide a value for the `pow_function_name` which doesn't have a default. So I'd rather prefer this choice to be explicit.

m2cgen/interpreters/code_generator.py
```diff
@@ -99,10 +99,11 @@ def prepend_code_lines(self, lines):
...
         self._write_to_code_buffer(f"{new_line.join(lines)}{new_line}", prepend=True)
 
-    # Following methods simply compute expressions using templates without changing result.
+    def infix_expression(self, left, right, op, wrap=False):
+        result = self.tpl_infix_expression(left=left, right=right, op=op)
+        return result if not wrap else f"({result})"
 
-    def infix_expression(self, left, right, op):
-        return self.tpl_infix_expression(left=left, right=right, op=op)
+    # Following methods simply compute expressions using templates without changing result.
```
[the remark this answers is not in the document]
izeigerman: Ah, right, I totally forgot the context behind that comment. Thanks for reminding me.

m2cgen/ast.py
```diff
@@ -7,6 +7,10 @@
 
 class Expr:
     output_size = 1
+
+    # The precedence of this expression. Higher value means higher precedence.
+    precedence = None
```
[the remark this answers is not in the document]
izeigerman: Ah, I'm somewhat unsure about this one. On one hand I agree that this simplifies the code quite a bit. On the other hand, I'm hesitant to introduce `precedence` for expressions to which this concept doesn't even apply. `None` in this case means "not applicable", while 0 represents an actual precedence. I feel like I'd like to preserve this distinction and keep it explicit to avoid unexpected behavior / unmet expectations in future. What do you think?

izeigerman:
@StrikerRUS I think I've addressed all your comments at this point except for `precedence = 0` being default for all expressions and the `PowExprFunctionMixin` as a default behavior (see my replies in respective threads). Thanks!

StrikerRUS:
> I'm going to regenerate code examples in a follow-up PR.

Let me automate the process of generating code examples. I believe we can make GitHub Actions generate code examples for each commit and pack them into a downloadable artifact. Then, if needed, we can download them and update `master` branch in a separate commit/PR. I can work on this if you don't mind.

m2cgen/ast.py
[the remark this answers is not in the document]
StrikerRUS: > I'm hesitant to introduce `precedence` for expressions to which this concept doesn't even apply.

Makes sense! After your clarification I'd like to preserve `None` as default value too 😃

m2cgen/interpreters/mixins.py
StrikerRUS: OK, fine with me.

tests/interpreters/test_java.py
```diff
@@ -288,10 +288,10 @@ def test_depth_threshold_with_bin_expr():
     expected_code = """
 public class Model {
     public static double score(double[] input) {
-        return (1.0) + ((1.0) + (subroutine0(input)));
+        return 1.0 + (1.0 + subroutine0(input));
     }
     public static double subroutine0(double[] input) {
-        return (1.0) + ((1.0) + (1.0));
+        return 1.0 + (1.0 + 1.0);
```
StrikerRUS, line 294: Sorry, I can't get it. Why do we have braces here (and in some other places like `var0 = (1.0 + 1.0);`)? Is it due to GE here?
```
wrap |= right_precedence is not None and right_precedence >= expr.precedence
```
https://github.com/BayesWitnesses/m2cgen/pull/511/files#diff-5a4f9c32da5612f71e48dad842ece55b645da127a61b6df8dee1c03ad3db76f1R208

StrikerRUS (approved):
Really great work! Thank you so much! I don't have any other suggestions except one non-blocking comment below. LGTM!

tests/interpreters/test_java.py
izeigerman: That's an excellent observation!
> Is it due to GE here?

You're absolutely right. Let me share some additional context.
The expression of kind:
```
1.0 + (1.0 + 1.0)
```
is produced as a result of the AST that looks like following:
```
ast.BinNumExpr(
    left=ast.NumVal(1.0),
    right=ast.BinNumExpr(
        left=ast.NumVal(1.0),
        right=ast.NumVal(1.0),
        op=ast.BinNumOpType.ADD
    ),
    op=ast.BinNumOpType.ADD
)
```
If the expression looked like this:
```
ast.BinNumExpr(
    left=ast.BinNumExpr(
        left=ast.NumVal(1.0),
        right=ast.NumVal(1.0),
        op=ast.BinNumOpType.ADD
    ),
    right=ast.NumVal(1.0),
    op=ast.BinNumOpType.ADD
)
```
You would've seen what you expected:
```
1.0 + 1.0 + 1.0
```
It may seem like this kind of difference is weird and it is for associative operations like addition or multiplication, but consider a slightly modified version of the first tree:
```
```
ast.BinNumExpr(
    left=ast.NumVal(1.0),
    right=ast.BinNumExpr(
        left=ast.NumVal(1.0),
        right=ast.NumVal(1.0),
        op=ast.BinNumOpType.SUB
    ),
    op=ast.BinNumOpType.SUB
)
```
This AST produces the following:
```
1.0 - (1.0 - 1.0)
```
Which makes a lot more sense now - omitting parenthesis would've been completely wrong in this case.

We can still get rid of the remaining parenthesis if we optimize our trees to look like 2nd tree rather than the 1st one for associative operation. Or we can introduce some sort of `is_associative` flag to our expressions which I found to be rather redundant. 

Let me know what you think.

Comment (izeigerman) [reply tests/interpreters/test_java.py:294]: Another alternative is to instead of having just `precedence` for each expression we can instead track `left_precedence` and `right_precedence` for each expression individually and get rid of `>=` comparison in the code snippet that you pointed out. 
For associative operators the `left_precedence` and `right_precedence` would have the same value, while for non-associative ones the `right_precedence` will be lower than the `left_precedence`.

Comment (izeigerman) [reply tests/interpreters/test_java.py:294]: Now when I thought about it more I really like the latter approach because it also generalizes well to right associative operators (like eg. `**` in Haskell). Let me know if you have any concerns and if not I'll update the implementation.

Comment (izeigerman) [reply tests/interpreters/test_java.py:294]: Ah, turns out my idea doesn't really cover scenarios like
```
1.0 - (1.0 + 1.0)
```
Please ignore the above comment.

StrikerRUS: Thank you so much for the detailed explanation with excellent examples! The situation is much clearer now for me. I totally support the idea of introducing some kind of associative rule tracking. But I'd like to propose making further updates in a follow-up PR. This PR is already quite big, self-contained and has already eliminate the great amount of excess braces.

izeigerman: Thanks! Merging this then.

PR #512, opened 2022-03-30
Generate artifact with code examples

StrikerRUS:
Use GitHub Actions to generate code examples for each commit and pack them into a downloadable artifact.
Artifact can be downloaded then from corresponding CI job page (can be found under the **Artifacts** heading after all jobs are finished).
image

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -9,6 +9,7 @@ if [[ $TEST == "API" ]]; then
   wget -q https://uploader.codecov.io/latest/linux/codecov -O codecov
   chmod +x codecov
   ./codecov -f coverage.xml -Z
+  python tools/generate_code_examples.py ./generated_code_examples
 fi
 
 if [[ $TEST == "E2E" ]]; then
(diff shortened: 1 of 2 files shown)


StrikerRUS:
Sorry, I'm not sure I've got your question. This PR is just proposing generation of updated content in the `generated_code_examples/` folder for each commit. If needed, that content can be downloaded as an artifact and used to update `generated_code_examples/` in the `master` branch. The aim of this PR is eliminate the need to setup environment and generate updated code examples locally.

StrikerRUS:
> I assume that the job that generates the code on each commit doesn't commit and/or pushes any changes made to that folder back to the repo.

Yeah, you're absolutely right! And I believe it shouldn't do this. Increased diff due to dozens of re-generated examples for a PR will require more maintainer's efforts to review and will shift reviewer's attention from actual changes.

I think it's better just make a separate PR that updates code examples from time to time.

> My question is whether we plan to remove the `generated_code_examples/` folder from the repo and only offer its content as an artifact produced by this job.

Oh, no, I don't have such plans! Newcomers unlikely would download weird zip archive just to get familiar with the package for the first time. GitHub has quite good syntax highlight for different programming languages - one click and potential user can easily inspect an example file if it's a part of the repo.

izeigerman (approved):
Awesome!

PR #519, opened 2022-05-08
Reduce the number of redundant parentheses in the generated code

StrikerRUS:
In the continuation of the discussion https://github.com/BayesWitnesses/m2cgen/pull/511#discussion_r838038667.

Introduce operator associativity to reduce the number of redundant parentheses even more.


m2cgen/interpreters/interpreter.py
```diff
@@ -203,9 +230,15 @@ def interpret_tanh_expr(self, expr, **kwargs):
...
 
-    def _wrap_infix_expr(self, expr, left_precedence, right_precedence):
+    def _wrap_infix_expr(self, expr, left_precedence, right_precedence, right_is_associative):
         wrap = left_precedence is not None and left_precedence > expr.precedence
-        wrap |= right_precedence is not None and right_precedence >= expr.precedence
+        wrap |= (
+            right_precedence is not None
+            and (right_precedence > expr.precedence
+                 or (right_precedence == expr.precedence
+                     and right_is_associative is not None
```
izeigerman, line 239: [Nitpick] can we reduce the condition to just `right_is_associative == False`? I realize it's a little counterintuitive.

m2cgen/interpreters/interpreter.py
```diff
@@ -203,9 +230,15 @@ def interpret_tanh_expr(self, expr, **kwargs):
         nested_result = self._do_interpret(expr.expr, **kwargs)
         return self._cg.function_invocation(self.tanh_function_name, nested_result)
 
-    def _wrap_infix_expr(self, expr, left_precedence, right_precedence):
+    def _wrap_infix_expr(self, expr, left_precedence, right_precedence, right_is_associative):
         wrap = left_precedence is not None and left_precedence > expr.precedence
-        wrap |= right_precedence is not None and right_precedence >= expr.precedence
+        wrap |= (
```
izeigerman, line 235: Maybe:
```
wrap |= right_precedence is not None and right_precedence > expr.precedence
wrap |= right_precedence is not None and right_precedence = expr.precedence and ...
```
for readability?

m2cgen/interpreters/interpreter.py
```diff
@@ -24,7 +24,15 @@ def interpret(self, expr):
...
         return None, kwargs
 
-    def _do_interpret(self, expr, to_reuse=None, left_precedence=None, right_precedence=None, **kwargs):
+    def _do_interpret(
+        self,
+        expr,
+        to_reuse=None,
+        left_precedence=None,
+        right_precedence=None,
+        right_is_associative=None,
```
izeigerman, line 33: I'm not sure about the naming here. Whether it's `right` or `left` doesn't really impact this particular parameter (unlike in case of `precedence`). I'd instead call it `parent_is_associative` or something.

Additionally, now I'm not quite sure about `*_precedence` names either. Maybe it'd be less confusing if we have something like this:
1. `parent_precedence` - pretty self-explanatory.
2. `is_right_from_parent` / `is_left_from_parent` - a boolean flag to indicate whether the current is the in the right  or left position.
3. `parent_is_associative` / `is_parent_associative` - the new parameter. 

What do you think?

m2cgen/interpreters/interpreter.py
StrikerRUS: Ah, indeed, thanks a lot for the tip!
But we should use `is False` instead of `== False` according to the
https://www.flake8rules.com/rules/E712.html

m2cgen/interpreters/interpreter.py
StrikerRUS: TBH, I intentionally hurt readability for the sake of code generation speed.
With current long complex expression we take benefits from the fact that `and` and `or` in Python are short-circuit operators:
https://docs.python.org/3/library/stdtypes.html#boolean-operations-and-or-not

With your advice about comparison with `False` above, I think, this expression becomes quite readable:
```python
    if right_precedence is not None:
        wrap |= (
            right_precedence > expr.precedence
            or (right_precedence == expr.precedence and right_is_associative is False)
        )
```
WDYT?

m2cgen/interpreters/interpreter.py
StrikerRUS: I like your proposal for `is_parent_associative` name.

But for `#1` and `#2` items I can't get the idea, unfortunately. Could you please provide a small code snippet with an example for these two params will be used in the wild?

izeigerman: Sure. Something like this:
```
def _wrap_infix_expr(...):
    if parent_precedence:
        if is_left_from_parent:
            return parent_precedence > expr.precedence
        return parent_precedence > expr.precedence or (parent_precedence == expr.precedence and is_parent_associative)
```

StrikerRUS:
Sorry for the delay!
I tried to address all review comments in f4b024b46e5dcfff25ae5437e9b13fa35d250458.

izeigerman (approved):
This looks great! Thanks a lot for addressing comments!

StrikerRUS:
> Do you have a strong preference to remove `... is True` part?

Kindly ping @izeigerman

izeigerman:
@StrikerRUS I'm so sorry, I totally missed your comment.

> Do you have a strong preference to remove ... is True part?

No, I get your point. No objections on my end.

StrikerRUS:
No problem at all!
Thanks a lot for your as always thoughtful review.

PR #544, opened 2022-09-15
Improve the development flow and extend the README with steps recommended before submitting a PR

izeigerman:


Makefile
```diff
@@ -1,21 +1,64 @@
...
+		chmod +x codecov
+
+run-codecov: download-codecov
+	./codecov -f coverage.xml -Z
 
 docker-build:
 	docker build -t $(DOCKER_IMAGE) .
 
 docker-test-all:
 	$(DOCKER_RUN_ARGS)
```
StrikerRUS, line 50: I believe this command isn't used anymore.
```suggestion
```
And this one is duplicated (can't make a GitHub suggestion):
https://github.com/BayesWitnesses/m2cgen/blob/2447b98553fa28c8ab1c00ab7a716c3c11721a76/Makefile#L66

Makefile
```diff
@@ -1,21 +1,64 @@
...
 
-docker-test-unit:
-	$(DOCKER_RUN_ARGS) bash -c "pytest -v --fast tests/ --ignore=tests/e2e/"
-
 docker-generate-examples:
 	$(DOCKER_RUN_ARGS) bash -c "python setup.py develop && python tools/generate_code_examples.py generated_code_examples"
```
StrikerRUS, line 52:
```suggestion
	$(DOCKER_RUN_ARGS) make generate-code-examples
```

README.md
```diff
@@ -23,6 +24,20 @@ Supported Python version is >= **3.7**.
 pip install m2cgen
 ```
 
+## Development
+The following system dependencies are required for API tests to run successfully:
+* gcc
+* cmake
+* libomp
```
StrikerRUS, line 31: Sorry, I can't find `cmake` and `libomp` in [our Dockerfile](https://github.com/BayesWitnesses/m2cgen/blob/master/Dockerfile). I guess they aren't required to run tests (especially, API ones).

m2cgen/VERSION.txt
```diff
@@ -1 +1 @@
-0.9.1
+0.10.1
```
StrikerRUS, line 1: Ah, I'm very sorry! I forgot to bump the version for development last time.
Thanks for fixing this!

requirements-test.txt
```diff
@@ -1,3 +1,12 @@
+# Other stuff
+numpy==1.23.3; python_version > "3.7"
+numpy==1.21.5; python_version <= "3.7"
+scipy==1.9.1; python_version > "3.7"
+scipy==1.7.3; python_version <= "3.7"
+
+# Cython
+Cython==0.29.32
```
StrikerRUS, line 8: Why do we need Cython again?
I remember we dropped it in https://github.com/BayesWitnesses/m2cgen/commit/f2c42fc897c942cdb447918332a131fc2dc95cd0.

StrikerRUS (changes requested):
Thanks a lot for re-organizing all this stuff! I left some comments below:

Makefile
izeigerman: Ah, good catch! Completely overlooked that, thanks!

README.md
izeigerman: Ah, seems like this is Mac OS specific. Let me clarify this in the README.

m2cgen/VERSION.txt
izeigerman: No problem at all!

README.md
izeigerman: Btw, all of these dependencies were needed to compile LightGBM during its installation.

izeigerman: For example I was getting an error like:
```
         raise Exception("\n".join((error_msg, LOG_NOTICE)))
      Exception: Please install CMake and all required dependencies first
      The full version of error log was saved into /Users/izeigerman/LightGBM_compilation.log
      [end of output]
 ```
 and further in compilation logs:
 ```
$ cat /Users/izeigerman/LightGBM_compilation.log
-- The C compiler identification is AppleClang 13.1.6.13160021
-- The CXX compiler identification is AppleClang 13.1.6.13160021
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Library/Developer/CommandLineTools/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Library/Developer/CommandLineTools/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at /opt/homebrew/Cellar/cmake/3.24.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find OpenMP_C (missing: OpenMP_C_FLAGS OpenMP_C_LIB_NAMES)
Call Stack (most recent call first):
  /opt/homebrew/Cellar/cmake/3.24.2/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
  /opt/homebrew/Cellar/cmake/3.24.2/share/cmake/Modules/FindOpenMP.cmake:545 (find_package_handle_standard_args)
  CMakeLists.txt:135 (find_package)
 ```
 Please note that I tried this on Mac with the M1 chip.

File: m2cgen/VERSION.txt (hunk 4 above)
Comment (izeigerman) [reply m2cgen/VERSION.txt:1]: I think we need a better process for this anyway.

File: requirements-test.txt (hunk 5 above)
Comment (izeigerman) [reply requirements-test.txt]: Seems like the same reason as in my other comment WRT LightGBM:
```
        File "lightning/setup.py", line 11, in cythonize_extensions
          raise ModuleNotFoundError(
      ModuleNotFoundError: Please install Cython in order to build a lightning from source.
      [end of output]
 ```
 It's actually falling back to compiling from source on my Mac with the M1 chip.

README.md
izeigerman: Deleted this block altogether for now since it's platform-specific.

requirements-test.txt
izeigerman: Please let me know whether you think we should keep it or drop it based on my use case.

README.md
StrikerRUS: Ah, you are right! Unfortunately, LightGBM doesn't provide wheels for macOS ARM architecture. This results in installation from sources when you `pip install lightgbm`.

requirements-test.txt
StrikerRUS: OK, I think we can make this dependency conditional then:
```suggestion
# Cython
Cython==0.29.32; sys_platform == 'darwin' and platform_machine == 'arm64'
```

izeigerman: Applied your change, thanks

izeigerman:
@StrikerRUS please let me know if there are any remaining comments

PR #546, opened 2022-09-26
Add support for Lua

ghost:
Lua is not as popular as Python or JavaScript. But Lua is the only language in which you can write fast and reliable API gateways, scripts for games, and many other cool things. It would be cool to add Lua support to m2cgen.


izeigerman:
Hey @pinktoxin ! Thank you so much for contributing this! So sorry about the delay with the review. I've just got back from vacation and will do my best to find time to review this PR soon. Appreciate your patience!

ghost:
Hello! Yesterday I tested the generated lua model in openresty and found a serious performance issue. My PR needs to be rewritten using one local lua table. I'll be back soon. Sorry for that

ghost:
The problem was that you can't create more than 200 local variables, but using global variables is not a solution because these variables in openresty (for example) live until the request ends. In Lua, these variables live until the program exits. That is why it is highly recommended not to use them.

ghost:
@izeigerman  Could you review the code?

PR #576, opened 2023-02-27
add code for multioutput regression

### LINKED ISSUE #559: Feature Request: support for multioutput regression
Nice library thanks!

Perhaps I missed something, but it looks like multi-output regression is unsupported? If so, is it on the roadmap? Happy to help if needed.

AaronDavidSchneider:
This PR adds the option for multioutput regression in XGBoost. This could work for other booster classes as well. However, there is no general attribute that informs about the number of targets. Thats why this solution is restricted to XGBoost.

Closes #559

### DIFF
--- a/m2cgen/assemblers/boosting.py
+++ b/m2cgen/assemblers/boosting.py
@@ -23,20 +23,24 @@ class BaseBoostingAssembler(ModelAssembler):
         self._is_classification = False
 
         model_class_name = type(model).__name__
-        if model_class_name in self.classifier_names:
-            self._is_classification = True
+        self._is_classification = model_class_name in self.classifier_names
+
+        if self._is_classification:
             if model.n_classes_ > 2:
                 self._output_size = model.n_classes_
+        else:
+            if getattr(self, "n_multioutput_targets", 1) > 1:
+                self._output_size = self.n_multioutput_targets
 
     def assemble(self):
-        if self._is_classification:
-            if self._output_size == 1:
+        if self._output_size > 1:
+            return self._assemble_multi_class_output(self._all_estimator_params)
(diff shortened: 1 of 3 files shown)


AaronDavidSchneider:
I hereby also add some tests. And here is another simple way to verify that it works:

```py
n_targets = 3
n_features = 3
n_test = 20
n_train = 20

X, y = datasets.make_regression(n_targets=n_targets, n_features=n_features, n_samples=n_train, random_state=1)

multi_class_model_params = {
    'n_estimators': 3,
    'max_depth': 2
}

model = XGBRegressor(**multi_class_model_params).fit(X, y)

code = m2cgen.export_to_python(model, function_name=f'predict')
with open('test_file.py', 'w') as f:
    f.write(code)

from test_file import predict

input = np.random.random((n_test, n_targets))

closenes = []
for i, input_i in enumerate(input):
    hardcoded = predict(input_i)
    apipred = model.predict(input_i.reshape((1, n_features)))
    closenes.append(np.allclose(apipred, hardcoded))

print(f'All close: {np.all(closenes)}, fraction of close: {np.sum(closenes)/n_test}')
```

Which works well for me.

I also ran the `make pre-pr` command, which reported `pass` for all tests but two on my MacBook. I don't know what happened to the two tests where it failed, but I suspect that it may be related to different versions of python packages being installed on my machine and the test machine (e.g., these exact same problems also occur on the master branch...).

Unfortunately, the docker command also fails on my machine, which does not seem to be caused by my changes but rather by the docker build command.

I hope that the CI will give more insight!

AaronDavidSchneider:
Kindly pinging @izeigerman and @StrikerRUS

